PR #2, opened 2022-04-22
DM-34315: Move rapid_analysis code to summit_utils and summit_extras

mfisherlevine:


python/lsst/summit/utils/utils.py
```diff
@@ -0,0 +1,392 @@
...
+from scipy.ndimage.filters import gaussian_filter
+import lsst.afw.detection as afwDetect
+import lsst.afw.math as afwMath
+import lsst.pipe.base as pipeBase
+import lsst.utils.packages as packageUtils
+
+from lsst.obs.lsst.translators.lsst import FILTER_DELIMITER
+from lsst.obs.lsst.translators.latiss import AUXTEL_LOCATION
+
+from astro_metadata_translator import ObservationInfo
```
isullivan, line 32: Is this the correct namespace for `astro_metadata_translator`?

isullivan (approved):
Looks good, with one minor question.

python/lsst/summit/utils/utils.py
mfisherlevine: Yup, it's actually standalone and lives outside the stack! Give it a go 🙂

PR #3, opened 2022-04-26
DM-34238: Make use of new butler init API

mfisherlevine:


python/lsst/summit/utils/bestEffort.py
```diff
@@ -69,18 +68,14 @@ class BestEffortIsr():
...
         self.log = logging.getLogger(__name__)
 
-        location = _repoDirToLocation(repodir)
-        collections = (LATISS_SUPPLEMENTAL_COLLECTIONS[location] if location in
-                       LATISS_SUPPLEMENTAL_COLLECTIONS.keys() else []) + LATISS_DEFAULT_COLLECTIONS
+        collections = LATISS_DEFAULT_COLLECTIONS
         self.collections = extraCollections + collections
         self.log.info(f'Instantiating butler with collections={self.collections}')
-        self.butler = dafButler.Butler(repodir, collections=self.collections,
+        self.butler = dafButler.Butler('LATISS', collections=self.collections,
```
czwa, line 78: Why not just use the `butlerUtils.makeDefaultLatissButler)` now that it exists?  Can't you pass the `run` to that method, since it already has the ability to make a writeable butler?

czwa (approved):
Small comment suggesting reuse of the improved `butlerUtils` method.

PR #6, opened 2022-04-29
DM-34606: Add test that DAF_BUTLER_REPOSITORY_INDEX point to file if set

mfisherlevine:


tests/test_butlerUtils.py
```diff
@@ -413,6 +413,13 @@ def test_makeDefaultLatissButlerRaiseTypes(self):
             with self.assertRaises(FileNotFoundError):
                 makeDefaultLatissButler()
 
+    def test_DAF_BUTLER_REPOSITORY_INDEX_value(self):
+        # If DAF_BUTLER_REPOSITORY_INDEX is truthy then we expect it to point
+        # to an actual file
+        repoFile = os.getenv('DAF_BUTLER_REPOSITORY_INDEX')
+        if repoFile:
+            self.assertTrue(os.path.isfile(repoFile))
```
ktlim, line 421: This is not true at the IDF.  I think you need `lsst.resources.ResourcePath(repoFile).exists()`.

ktlim (changes requested):

tests/test_butlerUtils.py
mfisherlevine: Good catch, thanks!

ktlim (approved):

PR #7, opened 2022-05-03
DM-34599: Create AllSkyCam RubinTV channel

mfisherlevine:


python/lsst/summit/utils/utils.py
```diff
@@ -26,6 +26,9 @@
 import lsst.pipe.base as pipeBase
 import lsst.utils.packages as packageUtils
 from lsst.daf.butler.cli.cliLog import CliLog
+import datetime
+from datetime import timedelta
```
czwa, line 30: Do you need to import this separately if you already have `datetime` imported?

czwa (approved):

python/lsst/summit/utils/utils.py
mfisherlevine: Thanks, I've fixed that, and also added the functions to `__all__` which I'd forgotten to do, and reformatted the `__all__` block to make future changes easier to diff.

PR #10, opened 2022-05-17
DM-34842: Add RSP as recognized location

mfisherlevine:


czwa (approved):
Do you need to add anything elsewhere for the case of `site == 'rsp'`, or do you simply need this to be a defined case so it doesn't raise a `ValueError`?  If it's the second, then I don't see any problems here.

mfisherlevine:
> Do you need to add anything elsewhere for the case of `site == 'rsp'`, or do you simply need this to be a defined case so it doesn't raise a `ValueError`? If it's the second, then I don't see any problems here.

Yup, just the latter 🙂

python/lsst/summit/utils/utils.py
```diff
@@ -454,7 +454,7 @@ def getSite():
     Returns
     -------
     location : `str`
-        One of ['tucson', 'summit', 'base', 'lsst-dev']
+        One of ['tucson', 'summit', 'base', 'rsp', 'lsst-dev']
```
ktlim, line 457: I might have called this `staff-rsp`; all of the deployments at the TTS, Summit, Base, and IDF are RSPs, which is a generic term.  Also note that this will move to USDF before August 15.

mfisherlevine: Good point about changing to `staff-rsp`. RE: USDF migration, as discussed with you last week, I aim to remove all of this site-specific stuff _long_ before then (it's only TTS that actually uses this at present).

I'll fixup and self merge the update.

mfisherlevine: Fixed on DM-34850.

PR #12, opened 2022-05-20
DM-34885: Write utility functions for calculating exposure sky position offsets

mfisherlevine:


python/lsst/summit/utils/utils.py
```diff
@@ -481,3 +485,118 @@ def getSite():
...
+        The second exposure.
+    useWcs : `bool`
+        Use the WCS for the ra/dec and alt/az if True, else use the nominal/
+        boresight values from the exposures' visitInfos.
+
+    Returns
+    -------
+    offsets : `lsst.pipe.base.Struct`
+        A struct contain the offsets:
+            - deltaRa
```
timj, line 551: Can you document the return types. Are they floats or Angle objects? I thought there was a documentation standard for Struct contents.

python/lsst/summit/utils/utils.py
```diff
@@ -481,3 +485,118 @@ def getSite():
...
+        dec1 = exp1.visitInfo.boresightRaDec[1]
+        dec2 = exp2.visitInfo.boresightRaDec[1]
+
+        p1 = exp1.visitInfo.boresightRaDec
+        p2 = exp2.visitInfo.boresightRaDec
+
+    angular_offset = p1.separation(p2).asArcseconds()
+    deltaPixels = angular_offset / pixScaleArcSec
+
+    ret = pipeBase.Struct(deltaRa=ra1-ra2,
```
timj, line 595: This wrapping needs to be handled or documented. Do you want 23 - 1 == 22 or -2? Should you use `.wrapNear(Angle(0.0))` on the result? (I haven't really checked). Do you handle this downstream?

python/lsst/summit/utils/utils.py
mfisherlevine: Angles, and yes, I found an example and have added the docs.

python/lsst/summit/utils/utils.py
mfisherlevine: So I tried doing `exp1 - exp2` and `exp2 - exp1` for two visits that are very closeby, and the results looked as I hoped. I then added `.wrapNear()` and the output is unchanged. The tests pass for both. Given that it seems to work both ways and is probably safer with it in, I'll leave it with that addition, and thanks for the tip, I didn't know about `.wrapNear()` 🙂

python/lsst/summit/utils/utils.py
```diff
@@ -481,3 +485,116 @@ def getSite():
...
+        The azimuth.
+
+    Notes
+    -----
+    Currently, as only the AuxTel is on-sky, the location is assumed to be
+    ``AUXTEL_LOCATION``. When these utils are extended to the main telescope,
+    this will need correcting.
+    """
+    skyLocation = SkyCoord(skyPos.getRa().asRadians(), skyPos.getDec().asRadians(), unit=u.rad)
+    altAz = AltAz(obstime=visitInfo.date.toPython(), location=AUXTEL_LOCATION)
```
timj, line 515: The irony being that `ObservationInfo` has all the information you need, including the Astropy AltAz object. More importantly, `VisitInfo` has the observatory location stored in it so you should not be guessing. `VisitInfo.getObservatory` returns `lsst.afw.coord.Observatory` which you should be able to transform into an astropy location.

timj (approved):
Please fix the AUXTEL_LOCATION hack. Other than that I've mostly looked at it before and it's fine.

python/lsst/summit/utils/utils.py
mfisherlevine: Good plan. Trivial to add to the code, torture to add to the test suite. Any idea how I'd add the location to the obsInfo/visitInfo I'm crafting by hand? I've tried a few things, but don't even really know where to look to work it out...

mfisherlevine: I tried adding some values for `telescope` to the base header dict but that didn't work.

timj: Are you making a VisitInfo from scratch or going via a header and ObservationInfo. The code to go from astropy location to Observatory is here: 

https://github.com/lsst/obs_base/blob/main/python/lsst/obs/base/makeRawVisitInfoViaObsInfo.py#L207-L212

mfisherlevine: Via a "header" and ObservationInfo (where the scare quotes denote it being a hand-made dict rather than an actual FITS header). You can see the code [here](https://github.com/lsst-sitcom/summit_utils/blob/6385fb5fe30719c8c124703a5339175cbb2488ee/tests/test_utils.py#L89)

mfisherlevine: Going between astropy location to Observatory isn't a problem, it's getting the test obsInfo/visitInfo to have it in it that's the problem.

mfisherlevine: What I meant by not knowing where to look is not really understanding what things to shove into the header to make it pop out in the obsInfo.

timj: The location comes from the OBSGEO-* FITS headers.

tests/test_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    """A test case for testing sky position offsets for exposures."""
+
+    def setUp(self):
+        camera = Latiss.getCamera()
+        self.assertTrue(len(camera) == 1)
+        self.detector = camera[0]
+
+        self.viMaker = MakeRawVisitInfoViaObsInfo()
+        self.mi = afwImage.maskedImage.MaskedImageF(0, 0)
+        self.baseHeader = dict(boresight_airmass=1.5,
```
timj, line 51: oh, this is a dict pretending to be an ObservationInfo? You need to create an astropy.coordinates.EarthLocation object and attach it to `location` in here.

python/lsst/summit/utils/utils.py
mfisherlevine: Right, but sticking values for those into the baseHeader dict doesn't get them into the output visitInfo, which is what I mean about not getting how to do this 😬

tests/test_utils.py
mfisherlevine: Ah, great, thanks!

tests/test_utils.py
```diff
@@ -0,0 +1,128 @@
...
+        header2['datetime_begin'] = astropy.time.Time(t2, format="isot", scale="utc")
+
+        header1['datetime_end'] = astropy.time.Time(t1+expTime, format="isot", scale="utc")
+        header2['datetime_end'] = astropy.time.Time(t2+expTime, format="isot", scale="utc")
+
+        obsInfo1 = makeObservationInfo(**header1)
+        obsInfo2 = makeObservationInfo(**header2)
+
+        vi1 = self.viMaker.observationInfo2visitInfo(obsInfo1)
+        vi2 = self.viMaker.observationInfo2visitInfo(obsInfo2)
```
timj, line 90: Use https://github.com/lsst/obs_base/blob/main/python/lsst/obs/base/makeRawVisitInfoViaObsInfo.py#L109

mfisherlevine: Huh? I am, on that very line, no? `self.viMaker` _is_ `MakeRawVisitInfoViaObsInfo`

mfisherlevine: but also there's no long any problems here.

timj: The `observationInfo2visitInfo` takes the ObservationInfo object you created above. The `__call__` method takes a header and first converts it to an ObservationInfo.

mfisherlevine: I don't understand what you're asking me to do. I added `location=AUXTEL_LOCATION` to the base header and everything works.

timj: You are right. Sorry. Trying to comment whilst in a meeting.

mfisherlevine: No probs. Merging 🎉

PR #20, opened 2022-10-04
DM-36400: Add OGA support to makeDefaultLatissButler

mfisherlevine:


python/lsst/summit/utils/butlerUtils.py
```diff
@@ -113,6 +113,9 @@ def makeDefaultLatissButler(*, extraCollections=None, writeable=False):
         Extra input collections to supply to the butler init.
     writable : `bool`, optional
         Whether to make a writable butler.
+    oga : `bool`, optional
+        Use the OGA repo instead of the main one. Needed to access embargoed
+    data.
```
czwa, line 118: I think this is supposed to be block indented.

czwa (approved):

PR #21, opened 2022-10-04
DM-36403: Add utility function for summing footprint fluxes

mfisherlevine:


python/lsst/summit/utils/utils.py
```diff
@@ -297,6 +299,80 @@ def detectObjectsInExp(exp, nSigma=10, nPixMin=10, grow=0):
...
+    """
+    median = 0
+    if subtractImageMedian:
+        median = np.nanmedian(parentImage.image.array)
+
+    # poor person's single dispatch
+    badTypeMsg = ("This function works with FootprintSets, single Footprints, and iterables of Footprints. "
+                  f"Got {type(footprints)}: {footprints}")
+    if isinstance(footprints, FootprintSet):
+        fps = footprints.getFootprints()
```
parejkoj, line 337: I'm not keen on `fps` here, but you've already got `footprints` as the arg name. It is an `if:elif:` block, so you could just re-assign to `footprints` in the places you needed to?

python/lsst/summit/utils/utils.py
```diff
@@ -297,6 +299,80 @@ def detectObjectsInExp(exp, nSigma=10, nPixMin=10, grow=0):
...
+    elif isinstance(footprints, Iterable):
+        if not isinstance(footprints[0], Footprint):
+            raise ValueError(badTypeMsg)
```
parejkoj, line 340: Is this a ValueError or a TypeError? Similarly below.

tests/test_utils.py
```diff
@@ -117,6 +121,45 @@ def test_getExpPositionOffset(self):
...
+        except FileNotFoundError:
+            raise unittest.SkipTest("Skipping tests that require the LATISS butler repo.")
+
+        cls.dataId = {'day_obs': 20200315, 'seq_num': 120, 'detector': 0}
+        cls.bestEffort = BestEffortIsr()  # should always succeed if makeDefaultLatissButler works
+        cls.exp = cls.bestEffort.getExposure(cls.dataId)
+
+    def test_fluxFromFootprint(self):
+        footPrintSet = detectObjectsInExp(self.exp)
+        allFootprints = footPrintSet.getFootprints()
```
parejkoj, line 140: This test doesn't need a butler at all, you could construct the image directly. For example, `afw/tests/test_footprint[1,2].py` place objects in an Image and make the footprints in place.

tests/test_utils.py
```diff
@@ -117,6 +121,45 @@ def test_getExpPositionOffset(self):
...
+        self.assertGreaterEqual(len(allFootprints), 3)
+        singleFootprint = allFootprints[0]
+        twoFootprints = allFootprints[0:2]
+
+        # check it can accept a footprintSet, and single and iterables of
+        # footprints
+        for obj in (footPrintSet, singleFootprint, twoFootprints):
```
parejkoj, line 147: I'd rather this not be in a loop: you have to do an `isinstance` as part of the test anyway. Pulling it out into individual blocks (possibly with a helper `_checkFluxFromFootprints` method) will make it easier to tell where a failure is.

tests/test_utils.py
```diff
@@ -117,6 +121,45 @@ def test_getExpPositionOffset(self):
...
+                self.assertEqual(len(fluxes), expectedLength)  # always one flux per footprint
+
+            self.assertIsInstance(fluxes, Iterable)
+            self.assertIsInstance(fluxes[0], np.floating)  # always returns floats
+
+        # ensure that subtracting the image median from fluxes leave image
+        # functionally unchanged
+        oldExpArray = copy.deepcopy(self.exp.image.array)
+        fluxes = fluxesFromFootprints(obj, self.exp, subtractImageMedian=True)
+        self.assertTrue(np.allclose(self.exp.image.array, oldExpArray))
```
parejkoj, line 160: Shouldn't this be identically equal? It shouldn't modify the image at all, I'd hope.

parejkoj (changes requested):
I think this could reasonably go into `afw/detection/utils.py` Any reason why you wouldn't want it here?

python/lsst/summit/utils/utils.py
mfisherlevine: Yup, sounds good, thanks.

tests/test_utils.py
mfisherlevine: Eurgh, it's floaty, and subtracting and adding back on a median in some cases, so I was just being careful. Perhaps I need to just do it differently though, and just subtract the median times the area in pixels from the footprint's flux  and never actually touch the pixels at all. That's probably safer, just a little more complex. Let me do it like that instead then.

python/lsst/summit/utils/utils.py
```diff
@@ -308,6 +310,80 @@ def detectObjectsInExp(exp, nSigma=10, nPixMin=10, grow=0):
...
+        The parent image.
+    subtractImageMedian : `bool`, optional
+        Subtract a whole-image median from each pixel in the footprint when
+        summing as a very crude background subtraction. Does not change the
+        original image.
+
+    Returns
+    -------
+    fluxes : `list` of `float`
+        The fluxes for each footprint.
```
parejkoj, line 334: Is there any value in returning a numpy array here instead?

python/lsst/summit/utils/utils.py
```diff
@@ -308,6 +310,80 @@ def detectObjectsInExp(exp, nSigma=10, nPixMin=10, grow=0):
...
+    """Calculate the flux from a footprint, given the parent image, optionally
+    subtracting a single value from each pixel as a very rough background
+    subtraction, e.g. the image median.
+
+    Parameters
+    ----------
+    footprint : `lsst.afw.detection.Footprint`
+        The footprint to measure.
+    parentImage : `lsst.afw.image.Image`
+        The parent image.
```
parejkoj, line 370:
```suggestion
        Image containing the footprint.
```

tests/test_utils.py
```diff
@@ -207,6 +213,59 @@ def test_quantiles(self):
             np.testing.assert_almost_equal(edges1, edges2, decimal=decimal)
 
 
+class ImageBasedTestCase(lsst.utils.tests.TestCase):
+    """A test case for testing sky position offsets for exposures."""
```
parejkoj, line 217: Are the docstring and/or class name right?

python/lsst/summit/utils/utils.py
```diff
@@ -308,6 +310,80 @@ def detectObjectsInExp(exp, nSigma=10, nPixMin=10, grow=0):
...
+
+    Returns
+    -------
+    flux : `float`
+        The flux in the footprint
+    """
+    if backgroundValue:  # only do the subtraction if non-zero for speed
+        xy0 = parentImage.getBBox().getMin()
+        return footprint.extractFluxFromArray(parentImage.array - backgroundValue, xy0)
+    return footprint.extractFluxFromImage(parentImage)
```
parejkoj, line 384: `extract` is now `compute` in all these function names.

tests/test_utils.py
```diff
@@ -207,6 +213,59 @@ def test_quantiles(self):
...
+        spans = afwGeom.SpanSet.fromShape(radius, afwGeom.Stencil.CIRCLE, offset=(44, 49))
+        footprint2 = afwDetect.Footprint(spans)
+        truth2 = spans.asArray() * image.array[34:41, 31:38]
+
+        allFootprints = [footprint1, footprint2]
+        footprintSet = afwDetect.FootprintSet(image.getBBox())
+        footprintSet.setFootprints(allFootprints)
+
+        # check it can accept a footprintSet, and single and iterables of
+        # footprints
```
parejkoj, line 244: Is it worth having an assertRaises for something that isn't one of those, just to make sure you didn't make your accept condition too broad?

tests/test_utils.py
```diff
@@ -207,6 +213,59 @@ def test_quantiles(self):
...
+        # check the list of footprints
+        fluxes = fluxesFromFootprints(allFootprints, image)
+        expectedLength = 2
+        self.assertEqual(len(fluxes), expectedLength)  # always one flux per footprint
+        self.assertIsInstance(fluxes, Iterable)
+        self.assertAlmostEqual(fluxes[0], np.sum(truth1))
+        self.assertAlmostEqual(fluxes[1], np.sum(truth2))
+
+        # ensure that subtracting the image median from fluxes leave image
+        # functionally unchanged
```
parejkoj, line 263: "functionally unchanged" or "completely unchanged"? The pixels should be untouched, no?

parejkoj (commented):
Handful of comments, most important being that you need to change `extract` to `compute` when calling the afw function.

python/lsst/summit/utils/utils.py
mfisherlevine: I'm happy with a list, personally. If the user needs, it can always be cast with `np.asarray` as needed.

tests/test_utils.py
mfisherlevine: Agree, docstring looks wrong. I'm going to leave fixing these to @fred3m though to save having too many cooks here.

python/lsst/summit/utils/utils.py
fred3m: I can't see the harm. I'll implement unless @mfisherlevine objects.

fred3m: I can't see the harm. I'll implement unless @mfisherlevine objects.

parejkoj (approved):

PR #22, opened 2022-10-04
DM-36449: Add check to test is var is there before deleting

mfisherlevine:


tests/test_butlerUtils.py
```diff
@@ -376,7 +376,8 @@ def test_dafButlerRaiseTypes(self):
         # If DAF_BUTLER_REPOSITORY_INDEX is not set *at all* then
         # using an instrument label raises a FileNotFoundError
         with unittest.mock.patch.dict('os.environ'):
-            del os.environ['DAF_BUTLER_REPOSITORY_INDEX']
+            if 'DAF_BUTLER_REPOSITORY_INDEX' in os.environ:  # can't del unless it's already there
```
ktlim, line 379: It's more compact to say `os.environ.pop('DAF_BUTLER_REPOSITORY_INDEX', None)`, but it might be less clear.

ktlim (approved):
This change alone is fine.  I haven't looked at any of the other code, though, and problems remain with packages in lsst_sitcom (e.g. missing summit_extras).

mfisherlevine:
> This change alone is fine. I haven't looked at any of the other code, though, and problems remain with packages in lsst_sitcom (e.g. missing summit_extras).

Yup, woke up to that fail on Jenkins. Should just be the fact that `summit_extras` wasn't in the table file for `rubintv_production` so I've added that to this same ticket number and am rerunning now...

tests/test_butlerUtils.py
mfisherlevine: Yeah, I think I like the explicit way in this instance.

PR #28, opened 2022-12-06
DM-37252: Move util from RubinTV to summit_utils

mfisherlevine:


python/lsst/summit/utils/utils.py
```diff
@@ -601,3 +606,46 @@ def getExpPositionOffset(exp1, exp2, useWcs=True, allowDifferentPlateScales=Fals
                           )
 
     return ret
+
+
+def starTrackerFileToExposure(filename, logger=None):
```
isullivan, line 611: This should probably be added to __all__

isullivan (approved):
Main suggestion is to update `__all__`

python/lsst/summit/utils/utils.py
mfisherlevine: Good shout, yes, thanks 🙂

python/lsst/summit/utils/utils.py
```diff
@@ -601,3 +606,46 @@ def getExpPositionOffset(exp1, exp2, useWcs=True, allowDifferentPlateScales=Fals
...
+    logger : `logging.Logger`, optional
+        The logger to use for errors, created if not supplied.
+
+    Returns
+    -------
+    exp : `lsst.afw.image.Exposure`
+        The exposure.
+    """
+    if not logger:
+        logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
mfisherlevine: I'm pretty sure that's _always_ defined. Like, just open a python terminal and type `print(__name__)` and it should tell you that you're in `__main__`.

PR #31, opened 2023-01-10
DM-37218: Improve butlerUtils ability to deal with native butler types

mfisherlevine:


python/lsst/summit/utils/bestEffort.py
```diff
@@ -139,20 +139,26 @@ def _parseExpIdOrDataId(expIdOrDataId, **kwargs):
...
+        elif isinstance(expIdOrDataId, dict):
             _dataId = expIdOrDataId
             _dataId.update(kwargs)
+        elif isinstance(expIdOrDataId, dafButler.DimensionRecord):
+            return expIdOrDataId.dataId
+        elif isinstance(expIdOrDataId, dafButler.DataCoordinate):
+            return expIdOrDataId
         else:
-            raise RuntimeError(f"Invalid expId or dataId type {expIdOrDataId}")
+            raise RuntimeError(f"Invalid expId or dataId type {expIdOrDataId}: {type(expIdOrDataId)}")
```
TallJimbo, line 161: If you're interested, you can also write this as
```
match expIdOrDataId:
    case int() as expId:
        return {"expId": expId, **kwargs}
    case dafButler.DataCoordinate() as dataId:
        assert not kwargs
        return dataId
    case dafButler.DimensionRecord() as record:
        assert not kwargs
        return record.dataId
    case dict() as dataId:
        return {dataId, **kwargs}
raise RuntimeError(...)
```
I would at least recommend adding something like the `assert`s above to guard against `kwargs` being silently ignored unexpected.  Or you could use `DataCoordinate.standardize` to try to include them in the result in all cases.

python/lsst/summit/utils/butlerUtils.py
```diff
@@ -167,6 +168,33 @@ def datasetExists(butler, dataProduct, dataId, **kwargs):
...
+    dataId : `dict` or `lsst.daf.butler.DataCoordinate`
+        The updated dataId, with the same type as the input.
+    """
+    if isinstance(dataId, dafButler.DataCoordinate):
+        return dafButler.DataCoordinate.standardize(dataId, **kwargs)
+    elif isinstance(dataId, dict):
+        dataId.update(**kwargs)
+        return dataId
+    else:
+        raise ValueError(f"Unknown dataId type {type(dataId)}")
```
TallJimbo, line 195: `match` would work well here, too.

TallJimbo (approved):

python/lsst/summit/utils/bestEffort.py
mfisherlevine: I actually have no idea why that function is trying to stick the kwargs in, when I'm just passing `**kwargs` to all the butler calls directly. I've replaced this with the `match` syntax, but not done the asserts, and instead stopped the function taking kwargs in the first place. Elsewhere I am indeed now using `DataCoordinate.standardize` and as I change more of the util code to use native types, hopefully I can drop nonsense like this entirely, in time 🙂

python/lsst/summit/utils/butlerUtils.py
mfisherlevine: Done, thanks.

PR #32, opened 2023-01-25
DM-37700: Raise warnings when get fails for expected products

mfisherlevine:


python/lsst/summit/utils/bestEffort.py
```diff
@@ -27,7 +27,7 @@
 from lsst.daf.butler.registry import ConflictingDefinitionError
 
 from lsst.summit.utils.quickLook import QuickLookIsrTask
-from lsst.summit.utils.butlerUtils import getLatissDefaultCollections
```
czwa, line 30: There's the linter error (this is as close as I can put a comment I think), but otherwise this looks fine.

czwa (approved):

PR #35, opened 2023-02-04
DM-37309: Nominally bring NightReport channel online

mfisherlevine:
Realistically now mostly other urgent fix-ups which had to be done while I had this ticket open and deployed on the summit.

### DIFF
--- a/python/lsst/summit/utils/astrometry/plotting.py
+++ b/python/lsst/summit/utils/astrometry/plotting.py
@@ -24,6 +24,7 @@ import numpy as np
 import matplotlib.pyplot as plt
 from astropy.coordinates import Angle
 import astropy.units as u
+import copy
 
 from lsst.obs.lsst.translators.latiss import AUXTEL_LOCATION
 from .. import quickSmooth
@@ -59,14 +60,14 @@ def plot(exp, icSrc=None, filteredSources=None, saveAs=None,
         Smooth the image slightly to improve the visability of low SNR sources?
     """
     fig = plt.figure(figsize=(16, 16))
-    arr = exp.image.array
-    arr = np.clip(arr, clipMin, clipMax)
+    data = copy.deepcopy(exp.image.array)
+    data = np.clip(data, clipMin, clipMax)
     if doSmooth:
-        arr = quickSmooth(arr)  # smooth slightly to help visualize
-    plt.imshow(np.arcsinh(arr)/10,
+        data = quickSmooth(data)  # smooth slightly to help visualize
+    plt.imshow(np.arcsinh(data)/10,
                interpolation='None', cmap='gray', origin='lower')
 
-    height, width = exp.image.array.shape
+    height, width = data.shape
     leftFraction = .15  # fraction into the image to start the N/E compass
     rightFraction = .225  # fraction into the image to start the az/el compass
     fontsize = 20  # for the compass labels
--- a/python/lsst/summit/utils/astrometry/plotting.py
+++ b/python/lsst/summit/utils/astrometry/plotting.py
@@ -58,7 +58,7 @@ def plot(exp, icSrc=None, filteredSources=None, saveAs=None,
     doSmooth : `bool`, optional
         Smooth the image slightly to improve the visability of low SNR sources?
     """
-    plt.figure(figsize=(16, 16))
+    fig = plt.figure(figsize=(16, 16))
     arr = exp.image.array
     arr = np.clip(arr, clipMin, clipMax)
     if doSmooth:
@@ -139,3 +139,6 @@ def plot(exp, icSrc=None, filteredSources=None, saveAs=None,
     if saveAs:
         plt.savefig(saveAs)
     plt.show()
+
+    plt.gcf().clear()
+    del fig
--- a/python/lsst/summit/utils/astrometry/plotting.py
+++ b/python/lsst/summit/utils/astrometry/plotting.py
@@ -143,3 +143,4 @@ def plot(exp, icSrc=None, filteredSources=None, saveAs=None,
 
     plt.gcf().clear()
     del fig
+    del data
--- a/python/lsst/summit/utils/astrometry/plotting.py
+++ b/python/lsst/summit/utils/astrometry/plotting.py
@@ -32,7 +32,7 @@ from .. import quickSmooth
 # TODO: Add some of Craig's nice overlay stuff here
 
 def plot(exp, icSrc=None, filteredSources=None, saveAs=None,
-         clipMin=1, clipMax=1000000):
+         clipMin=1, clipMax=1000000, doSmooth=True):
     """Plot an exposure, overlaying the selected sources and compass arrows.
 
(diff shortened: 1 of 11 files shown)


python/lsst/summit/utils/utils.py
```diff
@@ -706,3 +708,51 @@ def getFieldNameAndTileNumber(field, warn=True, logger=None):
...
+    ----------
+    airmass : `float`
+        The airmass, greater than or equal to 1.
+
+    Returns
+    -------
+    correctionFactor : `float`
+        The correction factor to apply to the seeing.
+
+    Raises
```
TallJimbo, line 732: Need a `------` after this.  Same in `getFilterSeeingCorrection`.

python/lsst/summit/utils/nightReport.py
```diff
@@ -536,7 +537,8 @@ def getExposureMidpoint(self, seqNum):
             The midpoint, as an mjd float.
         """
         timespan = self.data[seqNum]['timespan']
-        return (timespan.begin.mjd + timespan.end.mjd) / 2
+        expTime = self.data[seqNum]['exposure_time']
+        return ((timespan.begin) + expTime / 2).to_datetime()
```
TallJimbo, line 556: If these are coming from the butler, I would expect the only difference between new and old to be for readouts with significant dark time (which IIRC is included in `timespan` but not `exposure_time`).  Is that actually why you changed this?

tests/test_nightReport.py
```diff
@@ -125,10 +125,23 @@ def test_rebuild(self):
...
+
+        midPoint = self.report.getExposureMidpoint(seqNumToUse)
+        record = self.report.data[seqNumToUse]
+
+        if expTime == 0:
+            self.assertGreaterEqual(midPoint, record['datetime_begin'].to_datetime())
+            self.assertLessEqual(midPoint, record['datetime_end'].to_datetime())
+        else:
+            self.assertGreater(midPoint, record['datetime_begin'].to_datetime())
+            self.assertLess(midPoint, record['datetime_end'].to_datetime())
```
TallJimbo, line 144: If these are butler dimension records or something derived from them, it'd be good to move away from the `datetime_begin` and `datetime_end` nomenclature in favor of `timespan.begin` and `timespan.end` when you get a chance.  The `datetime_*` forms were grandfathered in and require some special-casing, and while we may never get around to deprecating the, I'd prefer to see `timespan.*` in new code.

python/lsst/summit/utils/nightReport.py
```diff
@@ -107,9 +113,20 @@ def _load(self, filename):
...
+        self._expRecordsLoaded = loaded['_expRecordsLoaded']
+        self._obsInfosLoaded = loaded['_obsInfosLoaded']
+        dayObs = loaded['dayObs']
+        loadedVersion = loaded.get('version', 0)
+
         if dayObs != self.dayObs:
             raise RuntimeError(f"Loaded data is for {dayObs} but current dayObs is {self.dayObs}")
+        if loadedVersion < self._version:
+            self.log.critical(f"Loaded version is {loadedVersion} but current version is {self._version}."
+                              " Check carefully for compatibility issues/regenerate your saved report!")
```
TallJimbo, line 127: Given that you're using pickle, I think it's highly likely you'll get errors when reading incompatible versions long before you get here, since that's quite easy to get when the pickled data structures change.

python/lsst/summit/utils/utils.py
```diff
@@ -22,7 +22,7 @@
 import os
 import numpy as np
 import logging
-from scipy.ndimage.filters import gaussian_filter
+from scipy.ndimage import gaussian_filter
```
TallJimbo, line 25: Typo in commit message: "depracated"

python/lsst/summit/utils/astrometry/plotting.py
```diff
@@ -59,14 +60,14 @@ def plot(exp, icSrc=None, filteredSources=None, saveAs=None,
...
     """
     fig = plt.figure(figsize=(16, 16))
-    arr = exp.image.array
-    arr = np.clip(arr, clipMin, clipMax)
+    data = copy.deepcopy(exp.image.array)
+    data = np.clip(data, clipMin, clipMax)
     if doSmooth:
-        arr = quickSmooth(arr)  # smooth slightly to help visualize
-    plt.imshow(np.arcsinh(arr)/10,
+        data = quickSmooth(data)  # smooth slightly to help visualize
```
TallJimbo, line 66: If this is acting in-place (as the commit implies), maybe it'd be better not to do anything with the return value, just in terms of readability.

python/lsst/summit/utils/astrometry/utils.py
```diff
@@ -259,10 +259,11 @@ def runCharactierizeImage(exp, snr, minPix):
     charConfig.doApCorr = False
     charConfig.doDeblend = False
     charConfig.repair.doCosmicRay = False
-    charConfig.repair.doInterpolate = True
+
     charConfig.detection.minPixels = minPix
     charConfig.detection.thresholdValue = snr
     charConfig.detection.includeThresholdMultiplier = 1
+    charConfig.detection.reEstimateBackground = False
```
TallJimbo, line 266: I'm curious about the reasoning for this.

python/lsst/summit/utils/astrometry/plotting.py
```diff
@@ -143,3 +143,4 @@ def plot(exp, icSrc=None, filteredSources=None, saveAs=None,
 
     plt.gcf().clear()
     del fig
+    del data
```
TallJimbo, line 146: If this still isn't working, you might want to try `gc.collect()`.  If that doesn't work, @natelust is the the expert.

TallJimbo (approved):
This was very much a review of the trees, not the forest, but only minor comments.

python/lsst/summit/utils/nightReport.py
mfisherlevine: As discussed OOB, this was designed to have thee same value, but be a python datetime rather than a float.

python/lsst/summit/utils/astrometry/plotting.py
mfisherlevine: Thanks.

python/lsst/summit/utils/astrometry/utils.py
mfisherlevine: As discussed OOB, this was me not setting the detection background config correctly.

python/lsst/summit/utils/nightReport.py
mfisherlevine: As discussed OOB, I'm being careful to only pickle the scraped data and no code at all, so hopefully this will still work OK, but the warning seemed like a nice feature to add anyway, just in case.

python/lsst/summit/utils/astrometry/utils.py
mfisherlevine: And setting the detection background to the regular background config _does_ indeed work, just FYI.

PR #37, opened 2023-03-22
DM-38385: Write fastStarTracker analysis code

mfisherlevine:


python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -0,0 +1,551 @@
...
+        source.hsmCentroidY = moments.moments_centroid.y + bbox.minY - 1
+        source.localCentroidX = moments.moments_centroid.x - 1
+        source.localCentroidY = moments.moments_centroid.y - 1
+        sources.append(source)
+    return sortSourcesByFlux(sources)
+
+
+def checkResultConsistency(results, maxAllowableShift=5, silent=False):
+    """Check if a set of results are self-consistent.
+
```
cmsaunders, line 316: How likely is it that you will have issues with sources at the edge of detectors? Having the brightest star at an edge seems like it will have a lot of unwanted consequences if it scatters or drifts off the detector.

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -0,0 +1,551 @@
...
+        toPrint.append(f'Mean and stddev of fluxes = {fluxMean:.1f} ± {fluxStd:.1f} ADU')
+
+    if not silent:
+        for line in toPrint:
+            print(line)
+
+    return consistent
+
+
+def plotResults(results, sourceIndex=0, allowInconsistent=False):
```
cmsaunders, line 409: Something like `plotSourceVariation` or just `plotSource` would be more informative.

Also, if this is ever going to be used outside of notebooks, it would probably be helpful to actually return the plots.

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -0,0 +1,551 @@
...
+    fig = plt.figure(figsize=(10, 10))
+    ax4 = fig.subplots(1)
+
+    colors = np.arange(len(sources))
+    # gnuplot2 has a nice balance of nothing white, and having an intuitive
+    # progression of colours so the eye can pick out trends on the point cloud.
+    ax4.scatter([s.centroidX for s in sources], [s.centroidY for s in sources], c=colors, cmap='gnuplot2')
+    ax4.set_xlabel('x-centroid (pixels)', size=axisLabelSize)
+    ax4.set_ylabel('y-centroid (pixels)', size=axisLabelSize)
+    ax4.set_aspect('equal', 'box')
```
cmsaunders, line 466: Can you add a labelled colorbar or something else to indicate that color corresponds to the image?

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -0,0 +1,551 @@
...
+        source.localCentroidX = moments.moments_centroid.x - 1
+        source.localCentroidY = moments.moments_centroid.y - 1
+        sources.append(source)
+    return sortSourcesByFlux(sources)
+
+
+def checkResultConsistency(results, maxAllowableShift=5, silent=False):
+    """Check if a set of results are self-consistent.
+
+    For each image, check the number of detected sources are the same in each
```
cmsaunders, line 317: I would remove "For each image"

cmsaunders (commented):
This looks good to me. I just have some minor comments.

cmsaunders (approved):

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -0,0 +1,551 @@
...
+           )
+
+
+def tifToExp(filename):
+    """Open a tif image as an exposure.
+
+    Opens the file, sets a blank mask plane, and converts the data to
+    `np.float32` and returns an exposure, currently with no visitInfo.
+
+    Once we have a way of getting the expTime, set that.
```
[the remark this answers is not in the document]
mfisherlevine: Done, thanks. (DM-38422)

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -0,0 +1,551 @@
...
+                retStr += f'bbox = lsst.geom.{repr(v)}\n'
+            elif itemName == 'cutout':  # and don't spam the full moments
+                if v is None:
+                    retStr += 'cutout = None\n'
+                else:
+                    retStr += f'cutout = {type(v)}\n'
+        return retStr
+
+
+def analyzeFastStarTrackerImage(filename, boxSize, attachCutouts=True):
```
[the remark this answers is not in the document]
mfisherlevine: Yeah, good point, renamed 👍

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -0,0 +1,551 @@
...
+
+    # numbers from the hsm moments fit
+    hsmFittedFlux: float = np.nan
+    hsmCentroidX: float = np.nan
+    hsmCentroidY: float = np.nan
+    moments: galsim.hsm.ShapeData = None  # keep the full fit even though we pull some things out too
+
+    imageBackground: float = np.nan
+    imageStddev: float = np.nan
+    nSourcesInImage: int = np.nan
```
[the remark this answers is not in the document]
mfisherlevine: I definitely paused there, for exactly that reason. I then thought about calling it something like `siblings` because that would make it make sense (and that's how I expect it to be used) but it was too close to deblender terminology, plus that's begging for an off-by-one error too, so I stuck with this. But I certainly agree in principle.

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -0,0 +1,551 @@
...
+    consistent : `bool`
+        Are the results nominally consistent?
+    """
+    if isinstance(results, typing.ValuesView):  # in case we're passed a .values()
+        results = list(results)
+
+    if 0 in ([len(sourceSet) for sourceSet in results]):
+        if not silent:
+            print('Some results contain no sources. Results are therefore fundamentally inconsistent'
+                  ' and other checks cannot be run')
```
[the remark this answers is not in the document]
mfisherlevine: Ooof. It's a fair point. I guess that's technically consistent, but also useless. I've added a test for this above this line to deal with that though, thanks.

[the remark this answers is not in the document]
mfisherlevine: `ff1990a` in case you want to see that change.

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
mfisherlevine: Quite likely, and I totally agree that will cause problems. That was kind of the thrust of the consistency checking helper function. My thinking is that if it slips off the edge the you either get a massive jump in centroid, or a different number of source counts, or a huge jump in brightness, or something else. Hopefully that will help people find things when it's happening, but I don't think there's much that can be done about it, at this stage at least.

mfisherlevine: If it looks necessary, I think that coming back to all this code and implementing a better way of doing source association between the images would be a good plan, but 1) this was all written on a plane, and 2) all the data so far has no problems in source association, so I just didn't need to do it, and so didn't given the great urgency with all this stuff 🙂

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
mfisherlevine: Done, and updated the docs too, thanks.

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
mfisherlevine: Hmmm, interesting. It's the position in the sequence, let me take a look at that.

mfisherlevine: Done 🙂

PR #38, opened 2023-03-22
DM-38312: Get fastStarTracker solving

mfisherlevine:


python/lsst/summit/utils/astrometry/anet.py
```diff
@@ -138,21 +138,25 @@ def __init__(self,
...
 
-    def _writeConfigFile(self, wide):
+    def _writeConfigFile(self, wide, useGaia):
         """Write a temporary config file for astrometry.net.
 
         Parameters
         ----------
         wide : `bool`
             Is this a wide field image? Used to select the 4100 vs 4200 dir in
             the index file path.
```
arunkannawadi, line 148: Ignored if `useGaia` is ``True``.

mfisherlevine: Yup, thanks.

python/lsst/summit/utils/astrometry/utils.py
```diff
@@ -162,6 +162,11 @@ def patchHeader(header):
         # service
         header['CRPIX1'] = 1560.85
         header['CRPIX2'] = 1257.15
+    if header.get('SECPIX') == '0.67':
+        # use the fast camera chip centre until we know better
+        header['SECPIX'] = '0.6213'  # measured from a fit
```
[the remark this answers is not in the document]
mfisherlevine: This whole function is awful, but it's a poor-person's obs_package, essentially. I'm patching up header values here by hand as we learn more and more about the various starTracker cameras, and I'm determining which camera we're in by the plate scale. I just managed to get a measurement of the plate scale on the fast camera, and so I'm updating it here so that it makes it into the initial wcs with the right value. Once we have an obs package none of this will be necessary though.

arunkannawadi (approved):

PR #41, opened 2023-04-13
DM-38662: Add required lsst modules

mfisherlevine:


ktlim (approved):
I have verified that these are all needed, but I haven't verified that _only_ these are needed.

PR #42, opened 2023-04-13
DM-38700: Use afwImage.MaskedImage directly

timj:
Should not go through afwImage.maskedImage.MaskedImage.


mfisherlevine (approved):
No idea how you found that, or why I did that, but sure, looks good.

TallJimbo (approved):

PR #44, opened 2023-04-18
DM-38739: add a utility function for Tony-scaling 

hsinfang:


python/lsst/summit/utils/utils.py
```diff
@@ -820,3 +823,90 @@ def getFilterSeeingCorrection(filterName):
...
+    minVal : `float`
+        An integer smaller than the minimum value in the input data.
+    maxVal : `float`
+        An integer larger than the maximum value in the input data.
+    """
+    flatData = data.ravel()
+    size = flatData.size - np.count_nonzero(np.isnan(flatData))
+
+    minVal = np.floor(np.nanmin(flatData)) - 1.0
+    maxVal = np.ceil(np.nanmax(flatData)) + 1.0
```
mfisherlevine, line 854: Given you're using `floor` and `ceil` here, do we need the -1 and +1 on top of that? Shouldn't those be enough to guarantee covering the whole range?

python/lsst/summit/utils/utils.py
```diff
@@ -820,3 +823,90 @@ def getFilterSeeingCorrection(filterName):
...
+
+    This is equivalent to using the numpy function:
+        np.quantile(data, np.linspace(0, 1, nColors + 1))
+    but with a coarser precision, yet sufficient for our use case.
+    This implementation gives a speed-up.
+
+    Parameters
+    ----------
+    data : `np.array`
+        The input image data
```
mfisherlevine, line 877: I find it really hard to care about stuff like this, but if we're going by the DM style guide these should end with a period...

python/lsst/summit/utils/utils.py
```diff
@@ -820,3 +823,90 @@ def getFilterSeeingCorrection(filterName):
...
+
+
+def digitizeData(data, nColors=256):
+    """
+    Scale data into nColors using its cumulative distribution function.
+
+    Parameters
+    ----------
+    data : `np.array`
+        The input image data
```
mfisherlevine, line 901: And again.

tests/test_utils.py
```diff
@@ -187,6 +188,21 @@ def test_quickSmooth(self):
...
 
+class QuantileTestCase(lsst.utils.tests.TestCase):
+    def setUp(self) -> None:
+        return super().setUp()
+
+    def test_quantiles(self):
+        # We understand that our algorithm gives very large rounding error
+        # compared to the generic numpy method. But still test it.
+        np.random.seed(1234)
+        data = np.random.normal(100, 10, 100)
```
mfisherlevine, line 199: I think it might be worth having some 2d data in here (to make sure we don't break the ravelling) and maybe something with some bigger values, to make sure things will work when we have values up to 2^18 (which is the number of bits in our ADC).

mfisherlevine (approved):

python/lsst/summit/utils/utils.py
hsinfang: About adjusting the `maxVal`, the as-is `digitizeData` [implementation](https://github.com/lsst-sitcom/summit_utils/blob/f5620ec3fc723dbae77b9af24a9387d139838538/python/lsst/summit/utils/utils.py#L911) would fail in the case that the largest data point is an integer and falls on rightmost edge, because it'd then want the non-existing `(maxVal - minVal)` bin. Making `maxVal` a little larger does the trick. 

Not adjusting the `minVal` would work too, I think.

mfisherlevine: Ah, I see, that makes sense, thanks.

tests/test_utils.py
hsinfang: 👍  about 2d data. 

I hesitate about much bigger values. Two numbers are also of interest here:
- `(maxVal - minVal)` which is then the histogram range and we fix the bin size to 1 
- `nColors` which we'd just use 256 
Larger values means larger imprecision too. Sure we can adjust the test to allow that.  But I'm not sure how meaningful it really is. 

Thoughts? How about the updated numbers in the test?

mfisherlevine: Rather than iterating in the comments here, I just pushed a commit with the changes I was imagining - I know that's a little irregular, I hope that's OK, obviously feel free to edit it however you like!

I added: several different ranges and numbers of colours, and also set an element in the array to `nan` to check we're nan-safe (and therefore changed the `np.quantile` call to be `np.nanqualite`)

Let me know what you think (and if you don't like me pushing to your branch in general!)

hsinfang: Good ideas in testing a variety of parameters, in testing `nan`,  and so on.

The only thing is that I cannot convince myself that data of (mean, width)= (2, 0.1) would be meaningful here. So I adjusted it to be a little bigger, yet small for possible input images I think.    Also I'm adjusting the tolerance slightly depending on the inputs.

hsinfang: Most pixels on such image would be approximated into just 1 and 2 with the current implementation. Probably easier if we have some example images. I'll leave it to a future ticket.

PR #47, opened 2023-06-02
DM-32940: Use the new Butler.exists API, replacing datasetExists

timj:
Requires lsst/daf_butler#840

### DIFF
--- a/python/lsst/summit/utils/bestEffort.py
+++ b/python/lsst/summit/utils/bestEffort.py
@@ -27,7 +27,7 @@ import lsst.daf.butler as dafButler
 from lsst.daf.butler.registry import ConflictingDefinitionError
 
 from lsst.summit.utils.quickLook import QuickLookIsrTask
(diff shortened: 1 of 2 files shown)


timj:
@mfisherlevine I'm not entirely sure how you want to play this. Do you have users of your utility `datasetExists` function outside of this package? Are you going to be running this code with an old butler that doesn't have `.exists()`? The comment associated with the utility function indicated you were happy to remove the function when the new API exists.

mfisherlevine:
As discussed OOB, I'm not too worried about users, but there's one instance in `summit_extras` and two in `rubintv_production`, both of which I'm happy to fix up later, but adding a deprecation warning for now, so that stacks can be updated with the coming weekly and then make use of it, would be a much easier path to getting this removed. But yes, to the "should it be removed", very much so, the intent was indeed always to get rid of that (and in fact as much else in this file as possible tbh!)

andy-slac (approved):
Looks good.

PR #48, opened 2023-06-08
DM-39563: Test now raises FileNotFoundError

timj:
Previously the repository index look up system was not being caught properly and so the RuntimeError printed through from the Config code. Now we catch this problem internally and allow the requested repo alias to be used. This changes the exception type being caught in the test.

Requires lsst/daf_butler#849


andy-slac (approved):
Looks great.

PR #55, opened 2023-07-19
DM-40101: Use vcr to provide a mock EFD for testing

mfisherlevine:


tests/test_efdUtils.py
```diff
@@ -51,12 +54,24 @@
 except ImportError:
     HAS_EFD_CLIENT = False
 
+# Use mode="none" to run tests for normal operation.
+# To update files or generate new ones, make sure you have a working
+# connection to lsst-schema-registry-efd.ncsa.illinois.edu
```
hsinfang: Is this address true?  I'm a bit surprised to still see nasa dependency.

hsinfang: Oh nvm... I see you corrected this in a later commit.  I guess that commit could be squashed with the first one.

hsinfang (approved):
LGTM. Though I have zero knowledge about `vcr` and can't really provide any feedbacks about it. 

How did you generate the cassette files? Might be good to document some instructions for other developers how to maintain them going forwards?

PR #56, opened 2023-07-31
DM-39852: Have TMA events know their block numbers

mfisherlevine:


python/lsst/summit/utils/blockUtils.py
```diff
@@ -0,0 +1,438 @@
...
+            reason = row['reason']
+            statePoint = ScriptStatePoint(time=time, state=state, reason=reason)
+            statePoints.append(statePoint)
+
+        # likewise for the blockIds
+        if len(blockIds) > 1:
+            raise RuntimeError(f"Found multiple blockIds ({blockIds}) for {seqNum=}")
+        blockId = blockIds.pop()
+
+        blockInfo = BlockInfo(
```
isullivan, line 399: Do you need to preserve the relation between specific `salIndices` and `tickets` from the same row of the block? If so, that has been lost here.

tests/test_blockUtils.py
```diff
@@ -0,0 +1,198 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Test cases for utils."""
+
+import os
```
isullivan, line 24: Please alphabetize imports, and move the `lsst.utils.tests` import to the lsst section.

tests/test_blockUtils.py
```diff
@@ -0,0 +1,198 @@
...
+                f"{blockInfo.end}{DELIMITER}"
+                f"{blockInfo.salIndices}{DELIMITER}"
+                f"{blockInfo.tickets}{DELIMITER}"
+                f"{len(blockInfo.states)}"
+            )
+            # must store as string not tuple for json serialization
+            data[f'{block}{DELIMITER}{seqNum}'] = line
+
+    packageDir = getPackageDir("summit_utils")
+    dataFilename = os.path.join(packageDir, "tests", "data", "blockInfoData.json")
```
isullivan, line 94: It could be useful to provide `dataFilename` as a keyword argument, and default to this value if none is provided.

tests/test_blockUtils.py
```diff
@@ -0,0 +1,198 @@
...
+    with open(dataFilename, 'w') as f:
+        json.dump(data, f)
+
+
+class BlockParserTestCase(lsst.utils.tests.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        try:
+            cls.client = makeEfdClient()
+        except RuntimeError:
```
isullivan, line 104: Please add a comment describing when you would expect `makeEfdClient` to fail.

isullivan (approved):
A number of comments and suggestions, mostly for clarity.

python/lsst/summit/utils/blockUtils.py
```diff
@@ -0,0 +1,438 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import re
+import time
+import logging
+import pandas as pd
+import numpy as np
```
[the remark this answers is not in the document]
mfisherlevine: Soon I'll be adding pre-commit to this whole package I think, running `isort` and `black` on everything, so this will all be taken care of. I think this large ticket is actually the only blocker for doing that, so will just deal with this once this merges.

python/lsst/summit/utils/blockUtils.py
```diff
@@ -0,0 +1,438 @@
...
+    seqNum : `int`
+        The sequence number of the block.
+    begin : `astropy.time.Time`
+        The time the block execution began.
+    end : `astropy.time.Time`
+        The time the block execution ended.
+    salIndices : `list` of `int`
+        One or more SAL indices, relating to the block.
+    tickets : `list` of `str`
+        One or more SITCOM tickets, relating to the block.
```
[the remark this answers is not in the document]
mfisherlevine: These are actually specifically SITCOM tickets (this is how they're parsed out). I guess at some point, if it's necessary, I could change the parsing and try to suck all Jira ticket formats out, but so far I think it's only ever SITCOM type tickets, and that certainly makes the data parsing a lot easier too. If I do ever change that I'll be sure to update these docs, but as they are, this is actually correct.

python/lsst/summit/utils/blockUtils.py
```diff
@@ -0,0 +1,438 @@
...
+        self.log = logging.getLogger("lsst.summit.utils.blockUtils.BlockParser")
+        self.dayObs = dayObs
+
+        self.client = client
+        if client is None:
+            self.client = makeEfdClient()
+
+        t0 = time.time()
+        self.getDataForDayObs()
+        self.log.debug(f"Getting data took {(time.time()-t0):.2f} seconds")
```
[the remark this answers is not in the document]
mfisherlevine: Thanks - I think that's generally a very good point, but given this is only on init I think it's fine here, but in general I think that's an important point, yes.

python/lsst/summit/utils/blockUtils.py
```diff
@@ -0,0 +1,438 @@
...
+        t0 = time.time()
+        self.augmentData()
+        self.log.debug(f"Parsing data took {(time.time()-t0):.5f} seconds")
+
+    def getDataForDayObs(self):
+        """Retrieve the data for the specified dayObs from the EFD.
+        """
+        data = getEfdData(self.client, 'lsst.sal.Script.logevent_state', dayObs=self.dayObs)
```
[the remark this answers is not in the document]
mfisherlevine: In general, hard coding is obviously something to be very wary of, but as far as I know, this is the one and only place that the BLOCK info is recorded, and given this is inside of the `BlockParser`, that seems not just OK, but almost necessary. It's like if there was a utility function which was specifically for getting AuxTel elevation position, then it would be the right thing to do, in that case, to hard code the topic. I absolutely agree with the general point about hard coding strings though, but feel like in this case, it would just be confusing and add nothing (because nobody would ever change it, but it would suggest to them that they might then want to). @tribeiro can you just confirm that this is indeed the only place where I would ever need to look for TMA BLOCK info?

python/lsst/summit/utils/blockUtils.py
```diff
@@ -0,0 +1,438 @@
...
+
+        Returns
+        -------
+        data : `pandas.DataFrame`
+            The row data.
+        """
+        rowsForBlock = self.data[self.data['blockNum'] == block]
+        if seqNum is None:
+            return rowsForBlock
+        return rowsForBlock[rowsForBlock['blockSeqNum'] == seqNum]
```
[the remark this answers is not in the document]
mfisherlevine: I think I actually want an empty dataframe here if it's not in there. I'll document that behaviour though.

python/lsst/summit/utils/blockUtils.py
```diff
@@ -0,0 +1,438 @@
...
+        seqNum : `int`, optional
+            The sequence number, if specified, to get the row data for. If not
+            specified, all data for the specified block is returned.
+
+        Returns
+        -------
+        data : `pandas.DataFrame`
+            The row data.
+        """
+        rowsForBlock = self.data[self.data['blockNum'] == block]
```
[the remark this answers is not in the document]
mfisherlevine: Added a log warning (but it shouldn't be an error, I don't think).

python/lsst/summit/utils/blockUtils.py
mfisherlevine: I am almost certainly that I don't, no, but that's a good point, thanks!

tests/test_blockUtils.py
mfisherlevine: Thanks, I've moved the `import lsst.utils.tests` section, and will wait for `isort` to fix the rest.

tests/test_blockUtils.py
mfisherlevine: Good shout, thanks.

tests/test_blockUtils.py
mfisherlevine: That's a really good point, and I actually don't know when I would expect that. Maybe when running from some place that isn't one of our systems? But I think the more important point that you've made me realise here, is that I've not annotated this to use the `vcr` package here, so this will end up contacting an actual EFD, which is no good at all! I will get that part fixed up right now!

PR #60, opened 2023-09-01
Add support for all nan data when calculating quantiles and CDF

mfisherlevine:


python/lsst/summit/utils/utils.py
```diff
@@ -970,13 +979,16 @@ def getCdf(data, scale):
 
 def getQuantiles(data, nColors):
     """Get a set of boundaries that equally distribute data into
-    nColors intervals. The output can be used to make a colormap
-    of nColors colors.
+    nColors intervals. The output can be used to make a colormap of nColors
+    colors.
```
hsinfang, line 983: As you predicted I'm annoyed by the unnecessary docstring reformatting mixed in this commit.  The style fix should be in a separate commit.  When tooling does that by default, I might do `git add -p` so not to include them with the new changes.

hsinfang (approved):

python/lsst/summit/utils/utils.py
mfisherlevine: Yup, agreed, I dislike it probably almost as much as you do - fixed! At some point it might be good to have a chat about taking a deep breath and moving this and related packages to using pre-commit with black and isort etc...

hsinfang: 👍  from me on taking the plunge and converting all formatting and then we no longer worry about it

PR #62, opened 2023-09-29
DM-40984: Move vcr init to a test utility file

mfisherlevine:


python/lsst/summit/utils/tests/utils.py
```diff
@@ -0,0 +1,50 @@
...
+    all the relevant sites, and temporarily run with mode="all" via *both*
+    python/pytest *and* with scons, as these generate slightly different HTTP
+    requests for some reason. Also make sure to do all this at both the summit
+    and USDF. The TTS is explicitly skipped and does not need to follow this
+    procedure.
+    """
+    packageDir = getPackageDir('summit_utils')
+    safe_vcr = vcr.VCR(
+        record_mode="none",
+        cassette_library_dir=os.path.join(packageDir, "tests", "data", "cassettes"),
```
timj, line 46: This is going to be a problem since we never install tests in docker containers. Putting this utility in public API means there is an expectation that people outside of the package should be able to use it. If you are wanting this to be private for your own tests then the normal approach is to move it into the tests directory itself and use `__file__` to find the test data.

mfisherlevine: Thanks. As discussed OOB, this should all be fixed now.

timj (approved):

PR #63, opened 2023-09-29
DM-40948: Add TMA eventMaker function for getting a specific event

mfisherlevine:


python/lsst/summit/utils/tmaUtils.py
```diff
@@ -1002,6 +1002,41 @@ def _mergeData(self, data):
 
         return merged
 
+    def getEvent(self, dayObs, seqNum):
+        """Get a specific event for a given dayObs and seqNum.
+
+        Repeated calls for the same ``dayObs`` will use the cached data if the
+        day is in the past, and so will be much quicker. If the ``dayObs`` is
+        the current day then the EFD will be queried for new data for each
+        call, so a call which returns ```None` on the first try might return an
```
jmeyers314, line 1011:
```suggestion
        call, so a call which returns ``None`` on the first try might return an
```

jmeyers314: And actually, might even just be `None`.  I'm fuzzy on the different use cases for single vs double `.

tests/test_tmaUtils.py
```diff
@@ -338,6 +341,25 @@ def test_helperFunctions(self):
...
+    @vcr.use_cassette()
+    def test_getEvent(self):
+        # test the singular event getter, and what happens if the event doesn't
+        # exist for the day
+        eventMaker = self.tmaEventMaker
+        events = eventMaker.getEvents(self.dayObs)
+        nEvents = len(events)
+
+        event = eventMaker.getEvent(self.dayObs, 0)
+        self.assertIsInstance(event, TMAEvent)
```
jmeyers314, line 353: Not sure if the API supports it, but could you add `self.assertEqual(event, events[0])` ?
And similar for `events[100]`?

jmeyers314 (approved):

python/lsst/summit/utils/tmaUtils.py
jmeyers314: Okay, that was supposed to be a single-backquoted None.  Not sure how to invoke that with github markdown...

mfisherlevine: I am equally fuzzy 😬 We can certainly both agree that three/one isn't the right format though 😛

tests/test_tmaUtils.py
mfisherlevine: Oh, it looks like past-Merlin actually did implement `__eq__` for these objects, so yes, I'll do that - nice!

PR #65, opened 2023-10-25
DM-41312: Fix blockInfo parsing for minimal data

mfisherlevine:


python/lsst/summit/utils/blockUtils.py
```diff
@@ -236,6 +236,15 @@ def augmentData(self):
         but is also much harder to maintain/debug, as the vectorized regexes
         are hard to work with, and to know which row is causing problems.
         """
+        if 'lastCheckpoint' not in self.data.columns:
+            self.log.warning("No lastCheckpoint column in data, can't parse block data")
```
isullivan, line 240: I would suggest `'lastCheckpoint'` to make it explicit that that is the name of the column and not a description, and state that empty columns are being added. Also, is it possible to print any sort of informational message about the data that was being attempted, such as an identifier or just the number of rows?

python/lsst/summit/utils/blockUtils.py
```diff
@@ -236,6 +236,15 @@ def augmentData(self):
         but is also much harder to maintain/debug, as the vectorized regexes
         are hard to work with, and to know which row is causing problems.
         """
+        if 'lastCheckpoint' not in self.data.columns:
```
isullivan, line 239: Do you also need this check in `augmentDataSlow`?

isullivan (approved):
Thanks for adding a comment explaining your use of `log.exception`.

python/lsst/summit/utils/blockUtils.py
mfisherlevine: Done. (When this happens there are almost always warnings about how little data is being found for the TMA as well btw)

python/lsst/summit/utils/blockUtils.py
mfisherlevine: I actually tested and didn't, but it was because I was getting zero rows of data, and thus the `iterrows()` there was meaning that line never got hit, so you're right, if I only got few but non-zero lines that would need be necessary! Thanks - added 👍

PR #72, opened 2023-11-28
DM-41885: Make TMAEvents hash so they work with sets

mfisherlevine:


python/lsst/summit/utils/tmaUtils.py
```diff
@@ -470,6 +470,22 @@ def __repr__(self):
             f" end={self.end!r}"
         )
 
+    def __hash__(self):
+        # deliberately don't hash the blockInfos here, as they are not
+        # a core part of the event itself, and are listy and cause problems
+        return hash((self.dayObs,
```
czwa, line 476: There can only be one source of these events, correct?

czwa (approved):
Looks fine as long as the answer to my question is "yes".

python/lsst/summit/utils/tmaUtils.py
mfisherlevine: I guess I'm not entirely sure what this means. This is a derived product, made from EFD data, and so for a fixed piece of EFD data and a given version of the generating code, events will be identical, but if EFD data were to change, or the generating code were to be updated, then the derived events could differ. I'm not sure I'm really answering the question, but I'm also not quite sure I understand it.

mfisherlevine: Any given person can recalculate these events though, so in that sense there are unlimited sources, but then I can't see how they'd need to `set` those together, so I doubt that's what you're asking?

PR #74, opened 2023-12-11
DM-42039: Add TMA mount tracking errors to TMA mount motion profile

craiglagegit:
This is the initial commit.  I think it is all working, but would like your comments, Merlin.  Here is a plot of a tracking event.
RubinTV_Tracking_Plot_20231123_431

### DIFF
--- a/python/lsst/summit/utils/tmaUtils.py
+++ b/python/lsst/summit/utils/tmaUtils.py
@@ -43,6 +43,7 @@ from .efdUtils import (getEfdData,
                        getDayObsForTime,
                        getDayObsStartTime,
                        getDayObsEndTime,
+                       clipDataToEvent,
                        )
 
 __all__ = (
@@ -305,6 +306,7 @@ def plotEvent(client,
                     " Pass in a figure with fig = plt.figure(figsize=(10, 8)) to avoid this warning.")
 
     fig.clear()
+    ax1p5 = None  # need to always be defined
     if event.type.name == 'TRACKING':
         ax1, ax1p5, ax2 = fig.subplots(3,
                                        sharex=True,
@@ -358,15 +360,19 @@ def plotEvent(client,
     ax2.xaxis.set_major_formatter(mdates.DateFormatter('%H:%M:%S'))
 
     if event.type.name == 'TRACKING':
-        azError = azimuthData['azError'].values
-        elError = elevationData['elError'].values
-        elVals = elevationData['actualPosition'].values
+        # XXX need to deal with clipping here
+        clippedAzimuthData = clipDataToEvent(azimuthData, event)  # returns a copy
+        clippedElevationData = clipDataToEvent(elevationData, event)  # returns a copy
+
+        azError = clippedAzimuthData['azError'].values
+        elError = clippedElevationData['elError'].values
+        elVals = clippedElevationData['actualPosition'].values
         if doFilterResiduals:
             # Filtering out bad values
-            azBadValues = filterBadValues(azError, maxDelta)
-            elBadValues = filterBadValues(elError, maxDelta)
-            azimuthData['azError'] = azError
-            elevationData['elError'] = elError
+            nReplacedAz = filterBadValues(azError, maxDelta)
+            nReplacedEl = filterBadValues(elError, maxDelta)
+            clippedAzimuthData['azError'] = azError
+            clippedElevationData['elError'] = elError
         # Calculate RMS
         az_rms = np.sqrt(np.mean(azError * azError))
         el_rms = np.sqrt(np.mean(elError * elError))
@@ -376,9 +382,9 @@ def plotEvent(client,
         image_az_rms = az_rms * np.cos(elVals[0] * np.pi / 180.0)
         image_el_rms = el_rms
         image_impact_rms = np.sqrt(image_az_rms**2 + image_el_rms**2)
-        ax1p5.plot(azimuthData['azError'], label='Azimuth error', c=lineColors[colorCounter])
+        ax1p5.plot(clippedAzimuthData['azError'], label='Azimuth tracking error', c=lineColors[colorCounter])
         colorCounter += 1
-        ax1p5.plot(elevationData['elError'], label='Elevation error', c=lineColors[colorCounter])
+        ax1p5.plot(clippedElevationData['elError'], label='Elevation tracking error', c=lineColors[colorCounter])
         colorCounter += 1
         ax1p5.axhline(0.01, ls='-.', color='black')
         ax1p5.axhline(-0.01, ls='-.', color='black')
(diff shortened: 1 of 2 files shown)


python/lsst/summit/utils/tmaUtils.py
```diff
@@ -116,9 +122,19 @@ def getTorqueMaxima(table):
...
+                                    prePadding=0,
+                                    postPadding=0,
+                                    doFilterResiduals=False,
+                                    maxDelta=0.1):
     """Get the data for the az/el telemetry topics for a given TMAEvent.
 
+    The error between the actual and demanded positions is calculated and added
+    to the dataframes in the az/elError columns. For TRACKING type events, this
+    error should be extremely close to zero, whereas for SLEWING type events,
+    this error
```
taranu, line 136: Unfinished sentence.

python/lsst/summit/utils/tmaUtils.py
```diff
@@ -150,11 +165,88 @@ def getAzimuthElevationDataForEvent(client, event, prePadding=0, postPadding=0):
...
 
 
-def plotEvent(client, event, fig=None, prePadding=0, postPadding=0, commands={},
-              azimuthData=None, elevationData=None):
+def filterBadValues(values, maxDelta=0.1):
+    """Filter out bad values from a dataset, replacing them in-place.
+
+    This function replaces non-physical points in the dataset with an
+    extrapolation of the preceding two values. No more than 3 successive data
+    points are allowed to be replaced. Minimum length of the input is 3 points.
```
taranu, line 187: I think anyone reading this in the future will wonder why do this at all instead of filtering/masking or highlighting the bad points in a plot? And I expect the answer will be some form of "we have a good reason to do it this way", but if you already have one or more in mind perhaps it's worth noting here.

python/lsst/summit/utils/tmaUtils.py
```diff
@@ -150,11 +165,88 @@ def getAzimuthElevationDataForEvent(client, event, prePadding=0, postPadding=0):
...
+        The maximum allowed difference between consecutive values. Values with
+        a difference greater than `maxDelta` will be considered as bad values
+        and replaced with an extrapolation.
+
+    Returns
+    -------
+    nBadPoints : `int`
+        The number of bad values that were replaced out.
+    """
+    # Find non-physical points and replace with extrapolation. No more than 3
```
taranu, line 203: Better make that a kwarg with a default 3 then, noting it's another 20+ lines before the value actually appears.

python/lsst/summit/utils/tmaUtils.py
```diff
@@ -150,11 +165,88 @@ def getAzimuthElevationDataForEvent(client, event, prePadding=0, postPadding=0):
...
+    log = logging.getLogger(__name__)
+
+    median = np.nanmedian(values)
+    # if either of the the first two points are more than maxDelta away from
+    # the median, replace them with the median
+    for i in range(2):
+        if abs(values[i] - median) > maxDelta:
+            values[i] = median
+            badCounter += 1
+            log.warning(f"Replacing bad value at index {i} with median value")
```
taranu, line 217: Seems like you may as well log the bad value and median while you're at it.

python/lsst/summit/utils/tmaUtils.py
```diff
@@ -150,11 +165,88 @@ def getAzimuthElevationDataForEvent(client, event, prePadding=0, postPadding=0):
...
+                consecutiveCounter += 1
+                badCounter += 1
+                log.warning(f"Replacing value at index {i} with {replacementValue}")
+                values[i] = replacementValue
+            else:
+                log.warning(f"More than 3 consecutive replacements at index {i}. Stopping replacements"
+                            " until the next good value.")
+        else:
+            consecutiveCounter = 0
+            replacementValue = (values[1] + values[0]) / 2.0
```
taranu, line 236: The comments lead me to believe this would be `(values[i] + values[i-1])/2.0`... this is just the average of the first two elements instead? Which you calculated outside of the loop and should not have changed.

But note even if it were the average of `values[i]` and `values[i-1]`, it's possible that `values[i-1]` was itself a bad replaced value, or a bad value that wasn't replaced at all if `consecutiveCounter` was too large.

python/lsst/summit/utils/tmaUtils.py
```diff
@@ -190,7 +282,12 @@ def plotEvent(client, event, fig=None, prePadding=0, postPadding=0, commands={},
     elevationData : `pd.DataFrame`, optional
         The elevation data to plot. If not specified, it will be queried from
         the EFD.
-
+    doFilterResiduals : 'bool', optional
+        Enables filtering of unphysical data points in the tracking residuals.
+    maxDelta : `float`, optional
+        The maximum difference between the model and the actual data, in
+        arcseconds, to allow before filtering the data point. Ignored if
+        ``filter`` is `False`.
```
taranu, line 290: if `doFilterResiduals` is `False`

python/lsst/summit/utils/tmaUtils.py
```diff
@@ -258,6 +363,55 @@ def tickFormatter(value, tick_number):
...
+        # returns a copy
+        clippedAzimuthData = clipDataToEvent(azimuthData, event, postPadding=TRACKING_RESIDUAL_TAIL_CLIP)
+        clippedElevationData = clipDataToEvent(elevationData, event, postPadding=TRACKING_RESIDUAL_TAIL_CLIP)
+
+        azError = clippedAzimuthData['azError'].values
+        elError = clippedElevationData['elError'].values
+        elVals = clippedElevationData['actualPosition'].values
+        if doFilterResiduals:
+            # Filtering out bad values
+            nReplacedAz = filterBadValues(azError, maxDelta)
```
taranu, line 376: Would it be useful to label/highlight/plot the replaced points in a different colour?

tests/test_tmaUtils.py
```diff
@@ -345,6 +346,57 @@ def test_helperFunctions(self):
         self.assertEqual(slews, foundSlews)
         self.assertEqual(tracks, foundTracks)
 
+    def test_filterBadValues(self):
```
taranu, line 349: I think you should have at least one test with a non-default maxDelta.

tests/test_tmaUtils.py
```diff
@@ -345,6 +346,57 @@ def test_helperFunctions(self):
...
+        values = np.array([1.0, 0.96, 1.0, 1.04, 0.95, 1.0, 1.05, 1.0, 1.05, 1.0, 0.95])
+        mean = np.mean(values)
+        nReplaced = filterBadValues(values)
+        self.assertEqual(nReplaced, 0)
+        self.assertEqual(np.mean(values), mean)
+
+        # test with one bad values
+        values = np.array([1.0, 0.96, 1.0, 1.04, 2.95, 1.0, 1.05, 1.0, 1.05, 1.0, 0.95])
+        nReplaced = filterBadValues(values)
+        self.assertEqual(nReplaced, 1)
```
taranu, line 361: Similarly, there ought to be at least one check that the expected elements were replaced with the expected values.

taranu (commented):
See comments.

python/lsst/summit/utils/tmaUtils.py
mfisherlevine: This was a good point. I'd meant to deal with this issue before, but didn't because the effort/reward seemed poor given where this goes in the codebase, but I ended up agreeing with you and putting it in, along with tests.

taranu (approved):
Looks good now, besides the one redundant line.

PR #75, opened 2023-12-11
DM-36211: Add PeekExposureTask

jmeyers314:
Adds `PeekExposureTask` as a solution to quickly find the brightest source (for observing scripts), and also quickly estimate PSF moments along xy, altaz, and equitorial axes.

### DIFF
--- a/python/lsst/summit/utils/assessPeekExposure.py
+++ b/python/lsst/summit/utils/assessPeekExposure.py
@@ -1,0 +1,275 @@
+# This file is part of summit_utils.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+
+import atexit
+from collections import namedtuple
+import logging
+import multiprocessing
+import shutil
+import time
+import os
+import warnings
+from pathlib import Path
+
+import lsst.afw.display as afwDisplay
+import matplotlib.pyplot as plt
+import numpy as np
+from astropy.table import Table
+from astropy.utils.exceptions import AstropyUserWarning
+from lsst.afw.geom import ellipses
+from lsst.daf.butler.datastore.cache_manager import DatastoreCacheManager
+from lsst.summit.utils.bestEffort import BestEffortIsr
+from lsst.summit.utils.peekExposure import PeekExposureTask
+from tabulate import tabulate
+from tqdm import tqdm
+
+# Set logger level to higher than CRITICAL to suppress all output
+silentLogger = logging.getLogger("silentLogger")
+silentLogger.setLevel(logging.CRITICAL + 1)
+
+
+# global vars for passing items from multiprocessing initializer to doWork.
+AssessPeekExposureGlobals = namedtuple(
+    "AssessPeekExposureGlobals", ["bestEffort", "pet", "display", "fig", "ax"]
+)
+
+
+SIGMA_TO_FWHM = 2.0 * np.sqrt(2.0 * np.log(2.0))
+AUXTEL_PIXEL_SCALE = 0.1  # arcsec/pixel
+
+
+# get one BestEffortIsr,afwDisplay per subprocess
+def initializePoolProcess():
+    """Initialize the multiprocessing pool process.
+
+    Sets up some variables we only want to create once per process, including
+    the BestEffortIsr, and afwDisplay + associated matplotlib figure objects.
+
+    Places these into a global variable, as that's the only way I know of to
+    effectively pass these to the doWork function.
+    """
+    global apeGlobals
+    fig, ax = plt.subplots(nrows=1, ncols=1, figsize=(9, 9), facecolor="k")
+    display = afwDisplay.Display(fig, backend="matplotlib")
+    display.scale("asinh", -1, 31)
+    petConfig = PeekExposureTask.ConfigClass()
+    pet = PeekExposureTask(config=petConfig, log=silentLogger, display=display)
+    bestEffort = BestEffortIsr(embargo=True)
+    apeGlobals = AssessPeekExposureGlobals(bestEffort, pet, display, fig, ax)
+
+
+# retrieve best-effort-isr and run PET on it
+def doWork(idx, row, doPlot):
+    """Run PeekExposureTask on a single exposure.
+
+    Parameters
+    ----------
+    idx : `int`
+        Index of the exposure in the table.
+    row : `astropy.table.Row`
+        Row of the table containing the exposure metadata.
+    doPlot : `bool`
+        Whether to plot the results.
+
+    Returns
+    -------
+    inTag : `str`
+        Input tag of the exposure.
+    outTag : `str`
+        Output tag of the exposure.
+    runtime : `float`
+        Runtime of the exposure in seconds.
+    exposureId : `int`
+        Exposure ID of the exposure.
+    """
+    global apeGlobals
+    bestEffort, pet, display, fig, ax = apeGlobals
+    exposureId = row["exposureId"]
+    dataId = {"instrument": "LATISS", "exposure": exposureId, "detector": 0}
+    exp = bestEffort.getExposure(dataId)
+    mode = "auto"
+    binSize = None
+    donutDiameter = None
+
+    # Until DM-41335 is fixed, we manually set mode for following exposures.
+    # Incorrectly labeled as photo
+    if exposureId >= 2022101200584 and exposureId <= 2022101200588:
+        mode = "donut"
+        donutDiameter = 1700
+    # Incorrectly labeled as giant donut
+    if exposureId >= 2022101200589 and exposureId <= 2022101200672:
+        mode = "photo"
+    if exposureId >= 2022101200673 and exposureId <= 2022101200676:
+        mode = "donut"
+        donutDiameter = 200
+    if exposureId >= 2022101200677 and exposureId <= 2022101200873:
+        mode = "photo"
+    if exposureId >= 2022101200874 and exposureId <= 2022101200875:
+        mode = "donut"
+        donutDiameter = 200
+    if exposureId >= 2022101200876 and exposureId <= 2022101201060:
+        mode = "photo"
+    if exposureId >= 2022101201061 and exposureId <= 2022101201062:
+        mode = "donut"
+        donutDiameter = 200
+    if exposureId >= 2022101201063 and exposureId <= 2022101201247:
+        mode = "photo"
+
+    t0 = time.time()
+    result = pet.run(
+        exp, mode=mode, doDisplay=doPlot, binSize=binSize, donutDiameter=donutDiameter
+    )
+    t1 = time.time()
+    runtime = t1 - t0
+
+    ax.text(
+        0.01,
+        0.97,
+        f"Runtime: {runtime:.2f} s",
+        c="w",
+        transform=ax.transAxes,
+        fontsize=12,
+    )
+
+    inTag = row["finalTag"]
+    trueX = row["centroid_x"]
+    trueY = row["centroid_y"]
+    foundX = result.brightestCentroid.x
+    foundY = result.brightestCentroid.y
+
+    if np.isfinite(trueX):
+        nx = (trueX - (result.binSize - 1) / 2) / result.binSize
+        ny = (trueY - (result.binSize - 1) / 2) / result.binSize
+        if doPlot:
+            display.dot("o", nx, ny, ctype=afwDisplay.MAGENTA, size=20)
+
+        if np.isfinite(foundX):
+            dist = np.sqrt((trueX - foundX) ** 2 + (trueY - foundY) ** 2)
+            if dist < 20:
+                outTag = "<2"
+            elif dist < 100:
+                outTag = "<10"
+            else:
+                outTag = ">10"
+        else:
+            outTag = "noStar"
+    else:
+        if np.isfinite(foundX):
+            outTag = "noTruth"
+        else:
+            outTag = "nothing"
+
+    if doPlot:
+        if np.isfinite(result.psfPixelShape.getIxx()):
+            distortion = ellipses.SeparableDistortionDeterminantRadius(result.psfPixelShape)
+            fwhm = SIGMA_TO_FWHM * distortion.getDeterminantRadius() * AUXTEL_PIXEL_SCALE
+        else:
+            fwhm = np.nan
+        display.show_colorbar(False)
+        ax.set_title(
+            f"{exposureId=}  {inTag=}  {outTag=}  {fwhm=:.2f} arcsec", color="w"
+        )
+        path = Path(args.plotdir)
+        path.mkdir(parents=True, exist_ok=True)
+        fn = f"test_pet_{idx:04d}_{exposureId:d}_{inTag}_{outTag}.png"
+        fig.savefig(path / fn)
+
+    return inTag, outTag, runtime, exposureId
+
+
+def main(args):
+    # Set up cache directory and register cleanup
+    defined, cacheDir = DatastoreCacheManager.set_fallback_cache_directory_if_unset()
+    if defined:
+        atexit.register(shutil.rmtree, cacheDir, ignore_errors=True)
+
+    # Loop through Merlin's curated data
+    path = Path(os.environ["SUMMIT_EXTRAS_DIR"]) / "data" / "qfm_baseline_assessment.parq"
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=AstropyUserWarning)
+        table = Table.read(path)
+    table["exposureId"] = table["day_obs"] * 100_000 + table["sequence_number"]
+    table = table[args.start:args.end]
+
+    intags = np.unique(table["finalTag"])
+
+    # Possible PET outcomes
+    outtags = [
+        "noStar",  # PET didn't find star, but catalog had one
+        "noTruth",  # No catalog star to compare to, but PET found one
+        "nothing",  # No catalog or PET star
+        "<2",  # PET brightest star agrees with catalog to better than 2 arcsec
+        "<10",  # agrees to 10 arcsec
+        ">10",  # disagrees by more than 10 arcsec
+    ]
+    results = {}
+    for intag in intags:
+        results[intag] = {}
+        for outtag in outtags:
+            results[intag][outtag] = []
+
+    if args.nproc == 1:
+        initializePoolProcess()
+        for idx, row in enumerate(tqdm(table, smoothing=0.05)):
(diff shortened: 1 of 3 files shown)


jmeyers314:
Started addressing flake8 failures reported here, then noticed that at some of my original formatting was apparently set by running first `isort` then `black`.  Am I wrong to think that black ought to produce acceptable whitespace-around-: and linebreak-before-binary-operator rules?  (Some of the other flake8 complaints are definitely legitimate though...)

jmeyers314:
Looks like the setup.cfg could set W503 under ignore as documented [here](https://developer.lsst.io/python/style.html#wrap-lines-before-binary-operators)?  And E203 is mentioned [here](https://developer.lsst.io/python/formatting.html#id1) 

Though I'm unsure if lsst-sitcom doesn't follow the same dev guide as lsst maybe?

mfisherlevine:
It does (currently, intend to) follow the same dev guide, yes. But it's not `black` and `isort`ed (yet), and there are places where `black` and `flake8` are mutually exclusive, in that `black` will autoformat things into a state that is not compatible with having an empty flake8 ignore-list, and the ignores picked by, for example, middleware packages, are not the same as those in pipelines, which is what this package follows. It's all a bit of a mess, but yes, the tl;dr is that `black` will, with the current configs, fight with `flake8`.

jmeyers314:
Well, easiest may be to just add W503 and E203 to https://github.com/lsst-sitcom/summit_utils/blob/main/setup.cfg#L4 (If I'm reading the dev guide right, then I think W503 should be added regardless of any potential black transition).  But I'm happy to do either.

python/lsst/summit/utils/assessPeekExposure.py
```diff
@@ -0,0 +1,200 @@
...
+    donutDiameter = None
+
+    # Incorrectly labeled as photo
+    if exposureId >= 2022101200584 and exposureId <= 2022101200588:
+        mode = "donut"
+        donutDiameter = 1500
+
+    # Incorrectly labeled as giant donut
+    if exposureId >= 2022101200589 and exposureId <= 2022101201247:
+        mode = "photo"
```
TallJimbo, line 73: I'm not sure how the "auto" resolution is done here, but I'm guessing it's based on FITS header info; if that's the case, these should get included as corrections in [obs_lsst](https://github.com/lsst/obs_lsst/tree/main/corrections), and the (at some point) you won't need this code anymore.

python/lsst/summit/utils/assessPeekExposure.py
```diff
@@ -0,0 +1,200 @@
...
+from tqdm import tqdm
+
+# Set logger level to higher than CRITICAL to suppress all output
+silent_logger = logging.getLogger("silent_logger")
+silent_logger.setLevel(logging.CRITICAL + 1)
+
+
+# get one BestEffortIsr,afwDisplay per subprocess
+def initializer():
+    global bestEffort, pet, display, fig, ax
```
TallJimbo, line 49: Hope you've got a good reason for using globals here.

Might be a bit better to have one class to package up all the global state, just so there's only actual one global variable.

python/lsst/summit/utils/assessPeekExposure.py
```diff
@@ -0,0 +1,200 @@
...
+    else:
+        if np.isfinite(found_x):
+            outTag = "noTruth"
+        else:
+            outTag = "nothing"
+
+    if doPlot:
+        if np.isfinite(result.psfShape.getIxx()):
+            distortion = ellipses.SeparableDistortionDeterminantRadius(result.psfShape)
+            fwhm = 2.35 * distortion.getDeterminantRadius() * 0.1
```
TallJimbo, line 120: I think I can guess what these random numbers mean, but it'd be better to make them module- or class-variable named constants.

python/lsst/summit/utils/assessPeekExposure.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def main(args):
+    # Set up cache directory and register cleanup
+    defined, cache_dir = DatastoreCacheManager.set_fallback_cache_directory_if_unset()
+    if defined:
+        atexit.register(shutil.rmtree, cache_dir, ignore_errors=True)
+
+    # Loop through Merlin's curated data
+    path = Path("/sdf/data/rubin/u/jmeyers3/tickets/DM-36211/summit_extras/data")
```
TallJimbo, line 142: This is not ideal, but again, I'm hoping you have a good reason for it?

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,772 @@
...
+        dtype=float,
+        default=16.0,
+        doc="Target number of pixels for a binned donut",
+    )
+    binSizeMax = pexConfig.Field(
+        dtype=int,
+        default=50,
+        doc="Maximum binning factor for donut mode",
+    )
+    # override installSimplePsf with donut kernel here.
```
TallJimbo, line 254: Is this a TODO entry that was suppose to happen on this branch?

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,772 @@
...
+        doc="PeekPhoto task",
+    )
+    spec = pexConfig.ConfigurableField(
+        target=PeekSpecTask,
+        doc="PeekSpec task",
+    )
+
+
+class PeekExposureTask(pipeBase.Task):
+    """
```
TallJimbo, line 463: Please give this a one-line summary, and the bulk of the docs will format a lot better if they're put in a `Notes` section instead of immediately after that.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,772 @@
...
+
+    For speed, many items in CharacterizeImageTask have been turned off,
+    including:
+        - PSF measurement (though not moment measurements)
+        - Deblending
+        - Repair
+        - Aperture correction
+        - Background estimation
+        - Noise replacement
+        - Writing to disk
```
TallJimbo, line 482: From this description it sounds like just running detection and measurement manually (I think that's all that's left) would be simpler, especially from a maintenance standpoint, as it will insulate you from upstream changes to a lot of things you don't care about.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,772 @@
...
+                - brightestCentroid : `lsst.geom.Point2D`
+                    Centroid of the brightest source (unbinned coords).
+                - brightestShape : `lsst.geom.Quadrupole`
+                    Shape of the brightest source (unbinned coords).
+                - psfShape : `lsst.geom.Quadrupole`
+                    Estimated PSF shape (unbinned coords).
+        """
+        # We're going to do binning and other manipulations, so just make a
+        # copy now and don't worry about modifying the original.
+        exposure = exposure.clone()
```
TallJimbo, line 595: There's no way binning is going to preserve the original image, so if you can make sure that happens first this deep-copy won't be necessary, and that might be useful if you're trying to squeeze performance.

TallJimbo (approved):
I'm hitting Approve because I don't know what the standards are for this package, or more importantly where the balance is between standing something up quickly and being able to maintain it, and so I'm happy to defer to the author on that.  There are issues here that I'd consider a maintenance problem if the code was going into `lsst_distrib`.

python/lsst/summit/utils/assessPeekExposure.py
jmeyers314: Yeah, I admit this is a bit of a hack, but globals are the only way I know of to get variables out of the multiprocessing.Pool initializer function.  I don't think this script should ever be imported elsewhere so hopefully the hack is contained at least.  Wrapping into a class is a good idea.

python/lsst/summit/utils/assessPeekExposure.py
jmeyers314: Yeah, there is a ticket already: https://jira.lsstcorp.org/browse/DM-41335?filter=13024.  I'll add a comment referencing that so we know when to delete this bit.

python/lsst/summit/utils/assessPeekExposure.py
jmeyers314: Not anymore.  Now that DM-34703 is merged I can use $SUMMIT_EXTRAS_DIR for at least a default value.

python/lsst/summit/utils/peekExposure.py
jmeyers314: More of: started working on it, decided I didn't need it, but didn't want to lose the partial progress.  So my tentative plan was to commit it, and then delete it so it at least lives on in git history.

python/lsst/summit/utils/peekExposure.py
jmeyers314: Sure, I'll look into that.  Using characterizeImage was helpful during development/optimization but you're right that what I've landed on here is much simpler.

jmeyers314: Okay, refactor worked.  Turns out the full set of subtasks is `installPsf`, `background`, `detection`, `measurement`, and possibly `catalogCalculation` (I didn't check if I actually care about the outputs of that last one).

I think I can make this a little tighter still by integrating the rebinning logic into the trimmed-down CharacterizeImageTask too, but that'll wait until tomorrow.

jmeyers314: Also, this refactor (which eliminates a whole round of duplicate measurement) + your don't-needlessly-clone-the-exposure suggestion look to make things go from ~1.0 s per ccd to ~0.5 s.  Thanks!

TallJimbo: :+1:   And I don't think you need `catalogCalculation`; it exists only to run the CModel-Psf "extendedness" star/galaxy classifier.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+        mode="auto",
+        donutDiameter=None,
+    ):
+        """
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Exposure at which to peek.
+        doDisplay : bool, optional
+            Display the exposure and sources?  Default False.
```
mfisherlevine, line 635: Maybe a tad picky, but might be worth noting here that a display is required to have been pass on init for this option.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+        if donutDiameter is None:
+            donutDiameter = self.getDonutDiameter(exposure)
+
+        mode, binSize, binnedSourceCat = self.runPeek(
+            exposure, mode, donutDiameter
+        )
+
+        table = self.transformTable(binSize, binnedSourceCat)
+
+        wgood = self.getGoodSources(mode, binnedSourceCat)
```
mfisherlevine, line 683: Maybe this is a convention I don't know of, but what's `w` here? _i.e._ why is this called `wgood`? I'd find it more readable if this was `goodSourceMask` or something like that.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+                - mode : str
+                    Peek mode that was run.
+                - binSize : int
+                    Binning factor used.
+                - binnedSourceCat : `lsst.afw.table.SourceCatalog`
+                    Source catalog from the binned exposure.
+        """
+        if mode == "auto":
+            # Note, no attempt to handle dispersed donuts.  They'll default to
+            # donut mode.
```
mfisherlevine, line 743: Just to say that is totally reasonable!

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+        match mode:
+            case "donut":
+                result = self.donut.run(exposure, donutDiameter)
+            case "spec":
+                result = self.spec.run(exposure)
+                if len(result.binnedSourceCat) == 0:
+                    self.log.warn("No sources found in spec mode.")
+                    if self.config.doPhotoFallback:
+                        self.log.warn("Falling back to photo mode.")
+                        result = self.photo.run(exposure, nobin=True)
```
mfisherlevine, line 760: Oh, is this the only reason `nobin` exists? If so then I think it's fine to leave it as-is, and just have the binning controlled by the binning factor then (as people can just use 1 for no binning). Out of interest though, why is this necessarily not binned? Because in this case we're known to be struggling, and therefore our best chance of winning is to go totally unbinned?

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+        return wgood
+
+    def getBrightest(self, binnedSourceCat, binSize, wgood):
+        """Find the brightest source in the catalog.
+
+        Parameters
+        ----------
+        binnedSourceCat : `lsst.afw.table.SourceCatalog`
+            Source catalog from the binned exposure.
+        binSize : int
```
mfisherlevine, line 845: Backticks on `int`

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+            Binning factor used.
+        wgood : `numpy.ndarray`
+            Boolean array indicating which sources are good.
+
+        Returns
+        -------
+        maxFluxIdx : int
```
mfisherlevine, line 852: And here. Unless this is not necessary for builtins maybe?! If so, I didn't know!

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+        return maxFluxIdx, b_centroid, b_shape
+
+    def getPsfShape(self, binnedSourceCat, binSize, wgood):
+        """Estimate the modal PSF shape from the sources.
+
+        Parameters
+        ----------
+        binnedSourceCat : `lsst.afw.table.SourceCatalog`
+            Source catalog from the binned exposure.
+        binSize : int
```
mfisherlevine, line 894: And again. It's systematic, so maybe you know something I don't 🙂

mfisherlevine:
Lots of quite picky comments from me, sorry. Generally, this package follows the DM dev guide as closely as is reasonably possible (and it's generally possible to follow it completely 🙂).

mfisherlevine:
Also, and this one is _really_ annoying, but this package is in camelCase, and this is bringing in a mix of snake and camel. Hopefully a quick find & replace can fix that up and not be too onerous?! 😬 Sorry!

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+
+    def run(self, exposure):
+        maskedImage = exposure.getMaskedImage()
+        median = np.nanmedian(maskedImage.image.array)
+        maskedImage.image.array -= median
+
+        return pipeBase.Struct(background=_MedianBackground(median))
+
+
+# Donut PSF; haven't actually tried this yet, but here's a start.
```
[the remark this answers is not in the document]
jmeyers314: Yep, haven't actually tried it.  At least partly after I realized the prior art was just installGaussianPsf and not installGenericPsf, and I didn't know how much more work would be required to extend on that.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+    def _doComputeBBox(self, position=None, color=None):
+        return Box2I(Point2I(-self.dimensions / 2), self.dimensions)
+
+    def _doComputeShape(self, position=None, color=None):
+        Ixx = self.R_outer**4 - self.R_inner**4
+        Ixx /= self.R_outer**2 - self.R_inner**2
+        return Quadrupole(Ixx, Ixx, 0.0)
+
+    def _doComputeApertureFlux(self, radius, position=None, color=None):
+        return 1 - np.exp(-0.5 * (radius / self.sigma) ** 2)
```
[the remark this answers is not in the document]
jmeyers314: These are all overrides, so the names are compelled.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+    """
+
+    ConfigClass = PeekPhotoTaskConfig
+    _DefaultName = "peekPhoto"
+
+    def __init__(self, config, **kwargs):
+        super().__init__(config=config, **kwargs)
+        self.makeSubtask("peek")
+
+    def run(self, exposure, nobin=False):
```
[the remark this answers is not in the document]
jmeyers314: To be honest, I don't remember now why I introduced a nobin option in the first place.  So I think I'll try removing it and seeing if anything breaks.

One addition that may be of interest is to have a config option to force a particular bin size.  I'll see that's straightforward to add.

python/lsst/summit/utils/peekExposure.py
jmeyers314: Haha.  This is channelling Josh from grad school where I extensively used the IDL `where` function and lazily set the return value to `w`.  Happy to use `goodSourceMask`.

python/lsst/summit/utils/peekExposure.py
jmeyers314: Oh, I think I remember now.  It's because having passed through self.spec.run, the exposure is already binned, so this is to prevent binning it a second time...

jmeyers314: Although, the clone() was supposed to prevent that?  I'll have to trace through to be certain...

python/lsst/summit/utils/peekExposure.py
jmeyers314: I'm sure I don't. :)

python/lsst/summit/utils/assessPeekExposure.py
```diff
@@ -0,0 +1,223 @@
...
+    with multiprocessing.Pool(args.nproc, initializer=initializer) as pool:
+        futures = []
+        for idx, row in enumerate(table):
+            futures.append(pool.apply_async(doit, (idx + args.start, row, args.plot)))
+
+        for row, future in zip(tqdm(table, smoothing=0.05), futures):
+            result, dt, inTag, outTag, idx, exposureId = future.get()
+            results[inTag][outTag].append((exposureId, dt))
+
+    from tabulate import tabulate
```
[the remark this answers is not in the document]
jmeyers314: cut-and-paste from your favorite GPT...  Will move.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+    )
+
+    def setDefaults(self):
+        super().setDefaults()
+        self.peek.detection.reEstimateBackground = False
+        self.peek.detection.doTempLocalBackground = False
+        self.peek.detection.minPixels = 40
+        self.peek.measurement.doReplaceWithNoise = False
+        self.peek.installPsf.fwhm = 10.0
+        self.peek.installPsf.width = 21
```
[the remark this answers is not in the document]
jmeyers314: Yeah, I moved common configs to the base.  We always override that in individual wrapper classes.

python/lsst/summit/utils/peekExposure.py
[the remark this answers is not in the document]
mfisherlevine: Gross! But defensible 🙂

python/lsst/summit/utils/peekExposure.py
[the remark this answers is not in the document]
mfisherlevine: Looks like it's used in the failover case when spectro fails and falls back to photo, just FYI.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -0,0 +1,1026 @@
...
+        doc="PeekPhoto task",
+    )
+    spec = pexConfig.ConfigurableField(
+        target=PeekSpecTask,
+        doc="PeekSpec task",
+    )
+
+
+class PeekExposureTask(pipeBase.Task):
+    """ Peek at exposure to extract summary statistics.
```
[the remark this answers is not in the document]
jmeyers314: How about:

```
Peek at exposure to quickly detect and measure both the brightest source 
in the image, and a set of sources representative of the exposure's overall 
image quality.
```

PR #76, opened 2023-12-11
DM-40512: fix bitrot in QuickLookIsrConnections

TallJimbo:


mfisherlevine (approved):
Thanks Jim - looks great, works great! Only thing I'm confused by is why the GitHub button is green when it needs rebasing (you got sniped by Eli). Surely it doesn't do that automatically now, does it? Do I need to enable something else GitHub to make it flag that?

TallJimbo:
It's just that I had already rebased by the time you finished your finished your review.

PR #77, opened 2023-12-14
DM-40517: add unit test for quickLook

isullivan:


mfisherlevine (commented):
This looks really great as a set of tests in general, and I think is really worth adding! There's one moderately-sized problem though: the issue before was in the actual command line task construction & execution (the connection magic), and this still doesn't get that tested with these changes. I reverted Jim's fix on DM-40512 to check, and the task then fails on the command line, but these tests still pass. I'm not sure exactly of the best way of testing this, but I thought that perhaps a `SimplePipelineExector` might be the way to go, as this will run it via the same machinery as the command line does.

isullivan:
I added a unit test that sets up a test butler with the necessary datasets, and which fails with the unmodified `quickLook`. I then fixed `quickLook` so that the new test passes, and that fixes the example `pipetask run` command at the USDF.

python/lsst/summit/utils/quickLook.py
```diff
@@ -224,7 +234,17 @@ def run(self, ccdExposure, *,
...
             self.log.info("Running with crosstalk correction")
 
-        if bfKernel:
+        if newBFKernel is not None:
+            try:
+                bfGains = newBFKernel.gain
+                isrConfig.doBrighterFatter = True
+                self.log.info("Running with new brighter-fatter correction")
+            except AttributeError:
+                bfGains = None
```
mfisherlevine, line 243: If a `newBFKernel` is supplied but we fail to pull the gains out then I'd think we'd probably want to at least log a warning here. What's your thinking on why this shouldn't be a `raise` though?

python/lsst/summit/utils/quickLook.py
```diff
@@ -233,20 +253,28 @@ def run(self, ccdExposure, *,
...
         isrConfig.doWrite = False
-        isrTask = IsrTask(config=isrConfig)
+        isrTask = self.isrTask(config=isrConfig)
         result = isrTask.run(ccdExposure,
                              camera=camera,
                              bias=bias,
                              dark=dark,
                              flat=flat,
-                             #  fringes=pipeBase.Struct(fringes=None),
+                             fringes=fringes,
```
mfisherlevine, line 262: Now that we're actually accepting the fringes if they're passed in, I think we should also have a block that tests if we got fringes supplied, and if so, sets the config option on to use them. We can then remove the `# TODO: deal with fringes here` comment too 🙂

python/lsst/summit/utils/quickLook.py
```diff
@@ -89,22 +89,32 @@ class QuickLookIsrTask(pipeBase.PipelineTask):
     ConfigClass = QuickLookIsrTaskConfig
     _DefaultName = "quickLook"
 
-    def __init__(self, **kwargs):
+    def __init__(self, isrTask=IsrTask, **kwargs):
         super().__init__(**kwargs)
+        # Pass in IsrTask so that we can modify it slightly for unit tests.
+        self.isrTask = IsrTask
```
mfisherlevine, line 95: This took me longer to understand than I care to admit, so I think a slightly more thorough comment might be useful here. Perhaps something along the lines of
```
        # Pass in IsrTask so that we can modify it slightly for unit tests.
        # Note that this is not an instance of the IsrTask class, but the class
        # itself, which is then instantiated later on, in the run() method,
        # with the dynamically generated config.
```

python/lsst/summit/utils/quickLook.py
mfisherlevine: I think just replacing that TODO with
```
        if fringes.fringes is not None:
            isrConfig.doFringe = True
            self.log.info("Running with fringe correction")
```
should do it.

tests/test_quickLook.py
```diff
@@ -0,0 +1,313 @@
...
+import lsst.ip.isr.isrMock as isrMock
+import lsst.ip.isr as ipIsr
+import lsst.pex.exceptions
+import lsst.pipe.base as pipeBase
+import lsst.pipe.base.testUtils
+from lsst.summit.utils.quickLook import QuickLookIsrTask, QuickLookIsrTaskConfig
+
+
+class QuickLookIsrTaskTestCase(unittest.TestCase):
+
```
mfisherlevine, line 41: Spurious newline.

tests/test_quickLook.py
```diff
@@ -0,0 +1,313 @@
...
+        deferredChargeCalib = "cpCtiCalib"
+        opticsTransmission = "transmission_optics"
+        strayLightData = "yBackground"
+        atmosphereTransmission = "transmission_atmosphere"
+        crosstalk = "crosstalk"
+        illumMaskedImage = "illum"
+        linearizer = "linearizer"
+        fringes = "fringe"
+        sensorTransmission = "transmission_sensor"
+        crosstalkSources = "isrOverscanCorrected"
```
mfisherlevine, line 136: Is that true?! I mean, I don't doubt you at all, but that's a horrible name mapping!

tests/test_quickLook.py
```diff
@@ -0,0 +1,313 @@
...
+
+        # Create a and populate a test butler for runQuantum tests.
+        self.repo_path = tempfile.TemporaryDirectory(ignore_cleanup_errors=True)
+        self.repo = butlerTests.makeTestRepo(self.repo_path.name)
+
+        # dataIds for fake data
+        butlerTests.addDataIdValue(self.repo, "instrument", instrument)
+        butlerTests.addDataIdValue(self.repo, "physical_filter", physical_filter, band=band)
+        butlerTests.addDataIdValue(self.repo, "detector", detector)
+        # butlerTests.addDataIdValue(self.repo, "detector", detector + 1)
```
mfisherlevine, line 152: Should this commented line be removed? Or was the plan to iterate over this option and check it raises for a detector id mismatch?

tests/test_quickLook.py
```diff
@@ -0,0 +1,313 @@
...
+        config.doCalculateStatistics = False
+
+        # Turn on all optional inputs
+        config.doAttachTransmissionCurve = True
+        config.doIlluminationCorrection = True
+        config.doStrayLight = True
+        config.doDeferredCharge = True
+        config.usePtcReadNoise = True
+        config.doCrosstalk = True
+        config.doBrighterFatter = False
```
mfisherlevine, line 261: I'm a little confused by this block in two small ways: it says to turn on all option inputs, but it doesn't include many (mostly the main ones, like bias/dark/flat etc), and also `config.doBrighterFatter` is set to `False` whereas all the others are `True`. Could you just clarify?

python/lsst/summit/utils/quickLook.py
isullivan: Maybe I took the statement "Should never fail" in the docstring of this task too literally! I'm happy to either log a warning or let it raise, your choice.

tests/test_quickLook.py
isullivan: Yeah: https://github.com/lsst/ip_isr/blob/113e1616294423797d5628c57463ee3dc7f1359c/python/lsst/ip/isr/isrTask.py#L123

tests/test_quickLook.py
isullivan: Good catch! I forgot to turn this back on after I got BF working.

python/lsst/summit/utils/quickLook.py
mfisherlevine: Ah, yeah, I see, I did wonder if it was something like that. I mean, it should never fail based if valid things are being passed in, but if, say, a calib was the wrong size or something truly awful like that then I think in _those_ cases it should still raise tbh, otherwise the behaviour is sort of ill-defined. Everything that's passed in should work, and if it doesn't then I think that's raise-time. All the other raises like that are implicit and come from isrTask raising though, so I think raising here is fine.

tests/test_quickLook.py
mfisherlevine: Wow, OK, thanks!

PR #79, opened 2024-01-30
DM-42719: Handle nan rotation angles in plotting

mfisherlevine:


python/lsst/summit/utils/astrometry/plotting.py
```diff
@@ -93,46 +93,49 @@ def plot(exp, icSrc=None, filteredSources=None, saveAs=None,
...
-              compassCenter[1],
-              compassSize*np.cos(rotpa),
-              compassSize*np.sin(rotpa),
-              color='green', width=20)
-    plt.text(compassCenter[0] + textDistance*np.cos(rotpa),
-             compassCenter[1] + textDistance*np.sin(rotpa),
-             'E',
-             color='green', fontsize=fontsize, weight='bold')
+
+    if not np.isnan(rotpa):  # need a rotation angle for the compass
```
taranu, line 97: Incredibly, I have a comment on a two-line PR... does it make sense for this and the check below to be `if np.isinfinite(rotpa)` instead? I suppose you don't need to check if it's between 0 and 2pi.

mfisherlevine: Errr, yeah, I guess. Honestly, I'd just never contemplated that they could be +/- inf, but I guess it doesn't hurt to use `isfinite` instead.

PR #81, opened 2024-02-15
DM-42836: Add calcPreviousDay convenience function

mfisherlevine:


python/lsst/summit/utils/efdUtils.py
```diff
@@ -498,10 +498,31 @@ def calcNextDay(dayObs):
         The next dayObs, as an integer, e.g. 20240101
     """
     d1 = datetime.datetime.strptime(str(dayObs), '%Y%m%d')
-    oneDay = datetime.timedelta(days=1)
+    negative = -1 if previous else 1
```
TallJimbo, line 501: Probably too late for this to be worthwhile, but an `offsetDayObs(dayObs, offsetInDays=1)` function that took an integer number of days would be even simpler to implement, a bit more general, and just as easy to use.

TallJimbo (approved):

python/lsst/summit/utils/efdUtils.py
mfisherlevine: No, that's a very fair point and minimal work. I've made that function, and just redirected the next/previous day functions to call that function with +/- 1 so there's still nicely readable wrappers for it.

PR #82, opened 2024-02-15
DM-41932: Add .associatedWith() method to TMAEvents

mfisherlevine:


cmsaunders (approved):
Looks good to me. My only comment is that "relatesTo" is quite vague. Would "associatedWith" or "includedIn" work? I don't understand `salIndex`, but I think it makes sense that a TMA event is "includedIn" a block or ticket.

PR #83, opened 2024-03-01
DM-42054: Remove limit of 1 for getCommandsDuringEvent

mfisherlevine:


python/lsst/summit/utils/efdUtils.py
```diff
@@ -661,3 +662,76 @@ def getTopics(client, toFind, caseSensitive=False):
...
+    Raises
+    ------
+    ValueError
+        Raise if there is already a command at a timestamp in the dictionary,
+        i.e. there is a collision.
+    """
+    if timeFormat not in ['pandas', 'astropy', 'python']:
+        raise ValueError(f"format must be one of 'pandas', 'astropy' or 'python', not {timeFormat=}")
+
+    commands = list(ensure_iterable(commands))
```
fritzm, line 705: Why "listify" here?  Can't the iterable be directly used in the `for` loop below?

python/lsst/summit/utils/tmaUtils.py
```diff
@@ -510,39 +540,29 @@ def getCommandsDuringEvent(client, event, commands=('raDecTarget'), log=None, do
...
-    # for triggering commands before the event
-
-    # TODO: DM-40100 Change this to always return a list of times, and remove
-    # warning about finding multiple commands. Remember to update docs and
-    # plotting code.
-    if log is None and doLog:
-        log = logging.getLogger(__name__)
-
-    commands = ensure_iterable(commands)
+    commands = list(ensure_iterable(commands))
```
fritzm, line 548: same question as above...

fritzm (approved):
LGTM at a coding level, one minor question.  You may still want Tiago's eyes for possible impact to the broader codebase?

python/lsst/summit/utils/efdUtils.py
mfisherlevine: I hate that too, but `ensure_iterable` will just return an exhaustible generator if it got one of those, and I hate accidentally consuming things when looping over them to find they're now gone, so just tend to do that for safety when I know the list is small, mainly just out of habit. It's not necessary here, but it's also probably less harmful than not doing it, on balance.

mfisherlevine:
Thanks. Should be fine though, this can't affect anything in Tiago-town (expect making it easier for him to use this code to debug things 🙂)

PR #85, opened 2024-03-01
DM-43137: Use dayObs as bind parameter instead of day_obs in queries

timj:


tests/test_blockUtils.py
```diff
@@ -67,8 +67,7 @@ def getBlockInfoTestTruthValues(dataFilename=None):
         The block info truth data.
     """
     if dataFilename is None:
-        packageDir = getPackageDir("summit_utils")
-        dataFilename = os.path.join(packageDir, "tests", "data", "blockInfoData.json")
+        dataFilename = os.path.join(TESTDIR, "data", "blockInfoData.json")
```
mfisherlevine, line 70: Is the package dir method no longer preferred? I don't see a problem with this switch, but just want to educate myself, as this method used to be the way to do it, and I'm curious why changed, and if I should change it everywhere.

tests/test_blockUtils.py
```diff
@@ -121,7 +119,7 @@ class BlockParserTestCase(lsst.utils.tests.TestCase):
     def setUpClass(cls):
         try:
             cls.client = makeEfdClient(testing=True)
-        except RuntimeError:
+        except Exception:
```
mfisherlevine, line 122: Ah, OK, that's interesting, I understand why you're getting that error, but I don't think this is the right fix. I think you should change the `"none"` on [this line](https://github.com/lsst-sitcom/summit_utils/blob/7c46278f81f0e7a3b30fbb3e7ebf3aa063c411f9/tests/utils.py#L44) to `"all"`, rerun once, commit the new data but not the none→all change, change that back, rerun the tests, and they should then pass (and actually run). Obviously run all the tests to capture all the data as this goes for all the `vcr` tests where you're changing the catch.

tests/test_efdUtils.py
```diff
@@ -64,7 +64,7 @@ class EfdUtilsTestCase(lsst.utils.tests.TestCase):
     def setUpClass(cls):
         try:
             cls.client = makeEfdClient(testing=True)
-        except RuntimeError:
+        except Exception:
```
mfisherlevine, line 67: As per other comment.

tests/test_ics.py
```diff
@@ -49,7 +49,7 @@ class M1M3ICSTestCase(lsst.utils.tests.TestCase):
     def setUp(cls):
         try:
             cls.client = makeEfdClient(testing=True)
-        except RuntimeError:
+        except Exception:
```
mfisherlevine, line 52: As per other comment.

tests/test_tmaUtils.py
```diff
@@ -227,7 +225,7 @@ class TMAEventMakerTestCase(lsst.utils.tests.TestCase):
     def setUpClass(cls):
         try:
             cls.client = makeEfdClient(testing=True)
-        except RuntimeError:
+        except Exception:
```
mfisherlevine, line 228: As per other comment.

tests/test_blockUtils.py
timj: I don't understand why this is needed though because the tests do pass on Jenkins. Does that mean they aren't really passing on Jenkins but some other code path is triggering a RuntimeError instead of the exception I get?

mfisherlevine: As discussed OOB, I think it's the 
```
try:
    import lsst_efd_client  # noqa: F401 just need to check this is available
except ImportError:
    HAS_EFD_CLIENT = False
```
`@unittest.skipIf(not HAS_EFD_CLIENT, "No EFD client available")` decorator that's skipping them as we don't have the client in the env 🙄

tests/test_blockUtils.py
timj: If you are looking for test data in a subdir relative to the test file, you should not use `getPackageDir` because you always want the test data files to go with the test python files regardless of how EUPS has been configured. This is future-proofed as well since it lets tests run in a future that doesn't always have EUPS.

You currently have to use `getPackageDir` in the package code to locate configs but at some point we really do need to move configs inside the python tree so we can use standard python package data to find the configs (which would let them work even without EUPS) but that is a big change that needs an RFC and changes to pex_config.

mfisherlevine: OK, sounds good, thanks.

PR #86, opened 2024-03-03
DM-43141: Fix StarTracker analysis to work with new headers

mfisherlevine:
Also fixes bitrot from config changes upstream and updates to astropy.

### DIFF
--- a/python/lsst/summit/utils/astrometry/utils.py
+++ b/python/lsst/summit/utils/astrometry/utils.py
@@ -150,19 +150,19 @@ def getAverageElFromHeader(header):
 def patchHeader(header):
     """This is a TEMPORARY function to patch some info into the headers.
     """
-    if header.get('SECPIX') == '3.11':
+    if header.get('CAMCODE') == 'GC102':  # regular aka narrow camera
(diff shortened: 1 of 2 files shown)


python/lsst/summit/utils/astrometry/utils.py
```diff
@@ -263,11 +263,13 @@ def runCharactierizeImage(exp, snr, minPix):
     charConfig.doMeasurePsf = False
     charConfig.doApCorr = False
     charConfig.doDeblend = False
+    charConfig.doMaskStreaks = False
```
isullivan, line 266: Probably not an issue for the star trackers, but note that some downstream code may fail if the STREAK mask is not present (I consider that a bug in our code). I believe you are already aware of this, but also note that streak masking is going to move to `detectAndMeasure` on difference images soon, and will not be part of the refactored `calibrateImageTask`.

isullivan (approved):

python/lsst/summit/utils/astrometry/utils.py
mfisherlevine: That's great, thanks 👍

PR #87, opened 2024-03-05
DM-43169: Create utility function to check for timebase errors

mfisherlevine:


python/lsst/summit/utils/auxtel/mount.py
```diff
@@ -0,0 +1,74 @@
...
+
+    cRIOtimestamps = mountPosition["cRIO_timestamp"]
+    if len(cRIOtimestamps) == 1:
+        log.warning(
+            f"cRIO_timestamp data had length 1 for {expRecord.obs_id}, so timebase errors are"
+            " impossible"
+        )
+        return False
+
+    return np.max(np.diff(cRIOtimestamps.values)) > maxDiff
```
isullivan, line 74: Should this be checking for `np.max(np.abs(...` instead?

python/lsst/summit/utils/auxtel/mount.py
```diff
@@ -0,0 +1,74 @@
...
+    timeseries.
+
+    Parameters
+    ----------
+    expRecord : `lsst.daf.butler.dimensions.DimensionRecord`
+        The exposure record to query.
+    client : `lsst_efd_client.efd_helper.EfdClient`
+        The EFD client to use.
+    maxDiff : `float`, optional
+        The maximum difference in cRIO timestamps to consider as a timebase
```
isullivan, line 46: Units are seconds? Also, is this the absolute value of the difference?

python/lsst/summit/utils/auxtel/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from .mount import *
```
isullivan, line 22: Maybe just import `hasTimebaseErrors` instead of `*`

isullivan (approved):
Please check whether you want the absolute value of the difference.

python/lsst/summit/utils/auxtel/mount.py
mfisherlevine: You know, kinda, but honestly, if time is running backwards then I think bets are so far off I'm happy to let it slide or explode in some other way.

python/lsst/summit/utils/auxtel/mount.py
mfisherlevine: Yup, will add comment on being seconds, though it's hard-coded in the cRIO, so it's somewhat irrelevant and would need re-crafting for other topics depending on how they're packed.

python/lsst/summit/utils/auxtel/mount.py
isullivan: OK, then I suggest you make this raise if the diff is negative.

PR #88, opened 2024-03-06
DM-42300: Handle day rollover in block parsing

mfisherlevine:


python/lsst/summit/utils/blockUtils.py
```diff
@@ -335,7 +344,12 @@ def getRows(self, block, seqNum=None):
         data : `pandas.DataFrame`
             The row data.
         """
-        rowsForBlock = self.data[self.data['blockNum'] == block]
+        # Because we query for a whole dayObs, but BLOCKs can overlap the day
+        # start/end, it's possible for the block's blockDayObs not to be the
+        # same as self.dayObs around the beginning or end of the day, so filter
+        # with an extra `& (self.data['blockDayObs'] == self.dayObs` when
+        # getting the relevant rows.
+        rowsForBlock = self.data[(self.data['blockNum'] == block) & (self.data['blockDayObs'] == self.dayObs)]
```
TallJimbo, line 352: It's more verbose, but I like using `numpy.logical_and` instead of `&` to avoid any chance of mixing up bitwise vs. logical in the rare scenarios where that matters.  But I think I may be in the minority.

TallJimbo (approved):
Looks good.  I hope it's not a problem that I didn't actually look at the YAML changes.

python/lsst/summit/utils/blockUtils.py
mfisherlevine: Yup, I like it - switched 🙂

mfisherlevine:
lol, yeah, never read the yaml changes 😄 There's a ticket to make them binary to stop people even considering it (and save space).

PR #91, opened 2024-03-06
DM-43199: Move to black and isort

mfisherlevine:


setup.cfg
```diff
@@ -1,7 +1,22 @@
...
+# N802: function name should be lowercase
+# N803: argument name should be lowercase
+# N806: variable in function should be lowercase
+# N812: lowercase imported as non lowercase
+# N813: camel case imported as lowercase
+# N815: mixed case imported as lowercase
+# N816: mixed case variable in class scope
+# W503: line break after binary operator, flake8 disagrees with PEP-8
+# W504: Line break occurred after a binary operator
+ignore = E133, E203, E226, E228, N802,
```
timj, line 17: I think E133 and E226  are impossible once you are enforcing black.

TallJimbo (approved):

setup.cfg
```diff
@@ -1,7 +1,22 @@
...
+         N803, N806, N812, N813, N815,
+         N816, W503, W504
```
timj, line 19: I have W503 set in middleware packages and that's the one listed in the developer guide.

setup.cfg
mfisherlevine: Removed.

setup.cfg
mfisherlevine: Unless there's a good reason to remove one of these, I'd probably lean towards keeping it, because it can be very annoying battling these sometimes. Is there a good reason to remove W504, do you think? Or is it more of a case of needing one, but not both?

timj: You can't have W503 and W504, you have to pick one for flake8 (they mean opposite things). We pick W504 because that matches black.

timj: and then I messed up my comment and said W504 by mistake (now fixed)

mfisherlevine: Well, they're rules to ignore, so I'd have thought you can pick all of them... but sure, will remove W504.

PR #93, opened 2024-03-11
DM-43255: Update analysis code and refactor

mfisherlevine:


python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -50,6 +54,29 @@
     "plotSourcesOnImage",
 )
 
+KNOWN_CAMERAS = ("regular", "wide", "fast")
```
taranu, line 57: Not narrow? Shouldn't this be KNOWN_CAMERA_TYPES then, if you plan to have another camera with `cameraType="regular"`?

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -50,6 +54,29 @@
...
+    doAstrometry: bool
+    cameraNumber: int
+    snr: float
+    minPix: int
+    brightSourceFraction: float
+    scaleError: float
+    doSmoothPlot: bool
+
+
+narrowCam = StarTrackerCamera("regular", "", "", True, 102, 5, 25, 0.95, 5, True)
```
taranu, line 76: Again I find it slightly confusing to have regular/narrow but I assume there is a reason for it.

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -80,6 +107,214 @@ def tifToExp(filename):
...
+    Returns
+    -------
+    isStreaming : `bool`
+        Whether the file is a streaming mode file.
+    """
+    # non-streaming filenames are like GC103_O_20240304_000009.fits
+    # streaming filenames are like GC103_O_20240304_000007_0001316.fits
+    # which is <camNum>_O_<dayObs>_<seqNum>_<streamSeqNum>.fits
+    # so 5 sections means streaming, 4 means normal
+    return len(os.path.basename(filename).split("_")) == 5
```
taranu, line 181: You can also do `os.path.basename(filename).count("_") == 5`, FYI.

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -451,36 +701,67 @@ def plotSourceMovement(results, sourceIndex=0, allowInconsistent=False):
...
     if not consistent and not allowInconsistent:
         checkResultConsistency(results.values(), silent=False)  # print the problem if we're raising
         raise ValueError("The sources were found to be inconsistent and allowInconsistent=False")
 
     sourceDict = {k: v[sourceIndex] for k, v in results.items()}
-    seqNums = list(sourceDict.keys())
+    frameNums = [s.frameNum for s in sourceDict.values()]
     sources = list(sourceDict.values())
 
+    allDayObs = set([s.dayObs for s in sources])
```
taranu, line 719: You can drop the list `[]` since it'll work fine with a generator, FYI.

taranu (commented):
See specific comments.

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
mfisherlevine: I don't think the `KNOWN_CAMERAS` needs renaming as they're physical cameras, not types, but yeah, this comment and all the others are absolutely right that I got part way through renaming `regular` to `narrow` but missed loads of places.

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
mfisherlevine: No no, you're right, this was also supposed to be renamed everywhere.

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
mfisherlevine: Nice, thanks, I do prefer that actually. Note the fence-post error there though - needs to be `os.path.basename(filename).count("_") == 4` for the same behaviour!

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
mfisherlevine: Ah thanks, I always forget that for some reason!

python/lsst/summit/utils/astrometry/fastStarTrackerAnalysis.py
```diff
@@ -50,6 +54,29 @@
...
+    snr: float
+    minPix: int
+    brightSourceFraction: float
+    scaleError: float
+    doSmoothPlot: bool
+
+
+narrowCam = StarTrackerCamera("regular", "", "", True, 102, 5, 25, 0.95, 5, True)
+wideCam = StarTrackerCamera("wide", "_wide", " wide", True, 101, 5, 25, 0.8, 5, True)
+fastCam = StarTrackerCamera("fast", "_fast", " fast", True, 103, 2.5, 10, 0.95, 60, False)
```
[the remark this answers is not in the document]
mfisherlevine: Yeah, totally agree, and I usually do use that. I'm also not using `slots` here either, which is a _really_ bad combination! This was because it was all growing very organically and in quite a panic a year ago when we first went on sky. Will tidy that up now 🙏

PR #94, opened 2024-03-12
DM-43269: Refactor StarTracker code

mfisherlevine:


python/lsst/summit/utils/utils.py
```diff
@@ -58,24 +58,33 @@
     "countPixels",
     "quickSmooth",
     "argMax2d",
+    "dayObsIntToString",
```
laurenam, line 61: Pre-existing, but wondering about the AUXTEL_LOCATION import dance?

laurenam (approved):

PR #102, opened 2024-05-09
DM-44104: Fix mypy errors

kkeijuro:
Check and solve and data typing code to be mypy compliant on summit_*

### DIFF
--- a/python/lsst/summit/utils/blockUtils.py
+++ b/python/lsst/summit/utils/blockUtils.py
@@ -471,7 +471,8 @@ class BlockParser:
             The events.
         """
         blockInfo = self.getBlockInfo(block, seqNum)
-        assert blockInfo is not None, f"No blockInfo found for {block=}, {seqNum=}"
+        if blockInfo is None:
+            return []
         begin = blockInfo.begin
         end = blockInfo.end
 
--- a/python/lsst/summit/utils/blockUtils.py
+++ b/python/lsst/summit/utils/blockUtils.py
@@ -332,7 +332,7 @@ class BlockParser:
         seqNums = seqNums.dropna()
         return sorted(set(seqNums))
 
-    def getRows(self, block: int, seqNum: int | None = None):
+    def getRows(self, block: int, seqNum: int | None = None) -> pd.DataFrame:
         """Get all rows of data which relate to the specified block.
 
         If the seqNum is specified, only the rows for that sequence number are
@@ -367,7 +367,7 @@ class BlockParser:
             return rowsForBlock
         return rowsForBlock[rowsForBlock["blockSeqNum"] == seqNum]
 
-    def printBlockEvolution(self, block: int, seqNum: int | None = None):
+    def printBlockEvolution(self, block: int, seqNum: int | None = None) -> None:
         """Display the evolution of the specified block.
 
         If the seqNum is specified, the evolution of that specific block
@@ -391,7 +391,7 @@ class BlockParser:
             blockInfo = self.getBlockInfo(block, seqNum)
             print(blockInfo, "\n")
 
-    def getBlockInfo(self, block: int, seqNum: int):
+    def getBlockInfo(self, block: int, seqNum: int) -> BlockInfo | None:
         """Get the block info for the specified block.
 
         Parses the rows relating to this block execution, and returns
@@ -412,7 +412,7 @@ class BlockParser:
         rows = self.getRows(block, seqNum=seqNum)
         if rows.empty:
             print(f"No {seqNum=} on dayObs={self.dayObs} for {block=}")
-            return
+            return None
 
         blockIds = set()
         tickets = set()
@@ -471,6 +471,7 @@ class BlockParser:
             The events.
         """
         blockInfo = self.getBlockInfo(block, seqNum)
+        assert blockInfo is not None, f"No blockInfo found for {block=}, {seqNum=}"
         begin = blockInfo.begin
         end = blockInfo.end
 
--- a/python/lsst/summit/utils/imageExaminer.py
+++ b/python/lsst/summit/utils/imageExaminer.py
@@ -25,7 +25,9 @@ __all__ = ["ImageExaminer"]
 import matplotlib
 import matplotlib.patches as patches
 import matplotlib.pyplot as plt
+from mpl_toolkits.mplot3d import Axes3D
 import numpy as np
+
 import scipy.ndimage as ndImage
 from matplotlib import cm
 from matplotlib.colors import LogNorm
@@ -324,7 +326,7 @@ class ImageExaminer:
         self.nSatPixInBox = countPixels(self.exp.maskedImage[self.starBbox], "SAT")
         return self.exp.image[bbox].array
 
-    def getMeshGrid(self, data: np.ndarray[int]) -> tuple[np.array, np.array]:
+    def getMeshGrid(self, data: np.ndarray[float]) -> tuple[np.ndarray[float], np.ndarray[float]]:
         """Get the meshgrid for a data array.
 
         Parameters
@@ -359,7 +361,7 @@ class ImageExaminer:
         center = np.array([xlen / 2, ylen / 2])
         # TODO: add option to move centroid to max pixel for radial (argmax 2d)
 
-        distances = []
+        distances: list[float] = []
         values = []
 
         # could be much faster, but the array is tiny so its fine
@@ -367,10 +369,11 @@ class ImageExaminer:
             for j in range(ylen):
                 value = self.data[i, j]
                 dist = norm((i, j) - center)
-                if dist > xlen // 2:
+                fDist = float(dist)
+                if fDist > xlen // 2:
                     continue  # clip to box size, we don't need a factor of sqrt(2) extra
                 values.append(value)
-                distances.append(dist)
+                distances.append(fDist)
 
         peakPos = 0
         amplitude = np.max(values)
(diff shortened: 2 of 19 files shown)


python/lsst/summit/utils/tmaUtils.py
```diff
@@ -908,6 +910,10 @@ def __setitem__(self, index: int, value: Any) -> None:
     def __len__(self) -> int:
         return len(self.keys)
 
+    def __iter__(self) -> Any:
+        for key in self.keys:
+            yield self.dictionary[key]
+
```
mfisherlevine, line 916: Out of interesting, what was necessitating this?

python/lsst/summit/utils/utils.py
```diff
@@ -373,7 +372,7 @@ def fluxesFromFootprints(
     )
     if isinstance(footprints, FootprintSet):
         footprints = footprints.getFootprints()
-    elif isinstance(footprints, Iterable):
+    elif isinstance(footprints, Sequence):
```
mfisherlevine, line 375: Why change `Iterable` to `Sequence` here?

python/lsst/summit/utils/type_utils.py
```diff
@@ -22,6 +23,14 @@ def dayObs(self) -> int:
...
         """Day of the observation."""
 
+    @property
+    def begin(self) -> Time:
+        """Day of the observation."""
+
+    @property
+    def end(self) -> Time:
+        """Day of the observation."""
+
```
mfisherlevine, line 33: I think this is a copy and paste error (which Bruno started). `seqNum` should read "The sequence number of the event within the day", and `begin` and `end` should read `The point at which the event begins` and `The point at which the event ends` respectively.

python/lsst/summit/utils/tmaUtils.py
```diff
@@ -1780,12 +1788,14 @@ def printTmaDetailedState(tma: TMAStateMachine) -> None:
...
-                f"InPosition: {p[f'{axis}InPosition']}"
-            )
+            f"{axis:>{axisPad}} - "
+            power_state = p[f"{axis}SystemState"]
+            motion_state = p[f"{axis}MotionState"]
+            if isinstance(power_state, PowerState):
+                f"Power: {power_state.name:>{powerPad}} "
+            if isinstance(motion_state, AxisMotionState):
+                f"Motion: {motion_state.name:>{motionPad}} "
+            f"InPosition: {p[f'{axis}InPosition']}"
```
mfisherlevine, line 1798: Two things:
1) please don't introduce snake_case to camelCase files
2) much more importantly, this totally changes the behaviour of the code!

- You have a floating f-string with no print statement at the start of the change, so that doesn't print.
- You've split one print statement into several, without setting `end=''` so this now prints on many lines where before it was a single line
- You finish with another unprinted f-string.

So I'd either: a) unpack everything, put in whatever asserts you need to to make mypy happy, and then call my print statement, or do it the way you're doing it, but put the missing print statements back in, and add `end=''` to all of them, and I _think_ that will give you the old behaviour back. In general, please be careful not to change the behaviour of things, especially things which are basically impossible to have decent tests for.

mfisherlevine: Before:
image
After:
image

python/lsst/summit/utils/imageExaminer.py
```diff
@@ -449,7 +451,7 @@ def plotRadialAverage(self, ax: matplotlib.axes.Axes | None = None) -> None:
 
         ax.plot(distances, values, "x", label="Radial average")
         if not fitFailed:
-            fitline = gauss(distances, *pars)
+            fitline = [gauss(distance, *pars) for distance in distances]
```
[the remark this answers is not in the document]
kkeijuro: Then X0 in gauss should also be an array also, isnt?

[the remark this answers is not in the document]
mfisherlevine: No, that's still a constant - numpy is fine with subtracting a constant from an array.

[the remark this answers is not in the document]
kkeijuro: Yes, you are right

python/lsst/summit/utils/imageExaminer.py
```diff
@@ -400,7 +402,7 @@ def radialAverageAndFit(self) -> None:
         # sort distances and values in step by distance
         d = np.array([(r, v) for (r, v) in sorted(zip(self.radialDistances, self.radialValues))])
         self.radii = d[:, 0]
-        values = d[:, 1]
+        values = list(d[:, 1])
```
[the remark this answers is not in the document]
kkeijuro: Not necessary, I dont know where the error came from...

python/lsst/summit/utils/nightReport.py
```diff
@@ -636,7 +638,8 @@ def plotPerObjectAirMass(
             obsTimes = [self.getExposureMidpoint(seqNum) for seqNum in seqNums]
             color = self.cMap[star].color
             marker = self.cMap[star].marker
-            plt.plot(obsTimes, airMasses, color=color, marker=marker, label=star, ms=10, ls="")
+            obsDates = date2num(obsTimes)
+            plt.plot(obsDates, airMasses, color=color, marker=marker, label=star, ms=10, ls="")
```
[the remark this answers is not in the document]
kkeijuro: Argument 1 to "plot" has incompatible type "list[datetime]"; expected "float | _SupportsArray[dtype[Any]] | _NestedSequence[_SupportsArray[dtype[Any]]] | bool | int | float | complex | str | bytes | _NestedSequence[bool | int | float | complex | str | bytes] | str"

python/lsst/summit/utils/utils.py
kkeijuro: Mypy complains about an Iterable not being indexable in next line:
    elif isinstance(footprints, Iterable):
        if not isinstance(footprints[0], Footprint): -> Should be a sequence

python/lsst/summit/utils/type_utils.py
kkeijuro: Yes, sorry

python/lsst/summit/utils/tmaUtils.py
kkeijuro: Yes, I forgot the print statements... Sorry! And snake_case...

python/lsst/summit/utils/utils.py
kkeijuro: Maybe changing the code changes code behaviour...

python/lsst/summit/utils/blockUtils.py
```diff
@@ -471,6 +471,7 @@ def getEventsForBlock(self, events: list[TMAEvent], block: int, seqNum: int) ->
             The events.
         """
         blockInfo = self.getBlockInfo(block, seqNum)
+        assert blockInfo is not None, f"No blockInfo found for {block=}, {seqNum=}"
```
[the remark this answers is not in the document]
mfisherlevine: Please add this log message back in:
`    self.log.info(f'No blockInfo found for {block=}, {seqNum=}')`

python/lsst/summit/utils/nightReport.py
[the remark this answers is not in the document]
mfisherlevine: Grrr, that's a matplotlib bug then, I guess, because it actually works just fine, and it's obviously doing something clever with them, and it should declare that I think. This change is fine though, thanks for that.

python/lsst/summit/utils/spectrumExaminer.py
```diff
@@ -128,7 +129,7 @@ def calcGoodSpectrumSection(self, threshold: int = 5, windowSize: int = 5) -> tu
         return (minPoint, maxPoint)
 
     def fit(self) -> None:
-        def gauss(x, a, x0, sigma):
+        def gauss(x: float, a: float, x0: float, sigma: float) -> float:
```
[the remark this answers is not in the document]
mfisherlevine: Here, and in the other file, this is an error: `TypeError: 'builtin_function_or_method' object is not subscriptable`.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -184,14 +185,14 @@ def __init__(self, size: float, outerRad: float, innerRad: float):
...
+    def __deepcopy__(self, memo: Any | None = None) -> DonutPsf:
         return DonutPsf(self.size, self.outerRad, self.innerRad)
 
-    def resized(self, width: float, height: float):
+    def resized(self, width: float, height: float) -> DonutPsf:
         assert width == height
         return DonutPsf(width, self.outerRad, self.innerRad)
 
-    def _doComputeKernelImage(self, position=None, color=None):
+    def _doComputeKernelImage(self, position: int | None = None, color: str | None = None) -> ImageD:
```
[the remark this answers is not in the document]
jmeyers314: Yep, for position it's a `Point2D`: https://github.com/lsst/afw/blob/b3a676b8dc86f8714a928f75276a329461bb25ed/include/lsst/afw/detection/Psf.h#L356.

And color is `lsst.afw.image.Color`, (https://github.com/lsst/afw/blob/b3a676b8dc86f8714a928f75276a329461bb25ed/include/lsst/afw/image/Color.h#L25).

python/lsst/summit/utils/tmaUtils.py
mfisherlevine: Huh, that feels a _little_ like a mypy bug, but I guess it's probably not. That's fine then, thanks for explaining.

python/lsst/summit/utils/utils.py
mfisherlevine: Ooooh, OK, then it probably _is_ a sequence rather than an iterable - shows what I know! Let's leave it as you have it then, thanks!

python/lsst/summit/utils/type_utils.py
mfisherlevine: Can you also change the error in the seqNum property I pointed out please?

python/lsst/summit/utils/tmaUtils.py
mfisherlevine: OK, you still haven't fixed this by the looks of things. Please let me know when you have and I'll re-re-review it.

PR #103, opened 2024-05-10
DM-43319: Fix warnings

mfisherlevine:


config/quickLookIsr.py
```diff
@@ -3,3 +3,4 @@
 config.doSaturationInterpolation = True
 config.overscan.fitType = "MEDIAN_PER_ROW"
 config.overscan.doParallelOverscan = True
+# config.brighterFatterMaxIter = 2  # Uncomment this to remove test warning
```
mfisherlevine, line 6: I think this should go back in - that's an acceptable warning to have.

python/lsst/summit/utils/butlerUtils.py
```diff
@@ -478,10 +478,12 @@ def _assureDict(dataId: dict | dafButler.dimensions.DataCoordinate | dafButler.D
     """
     if isinstance(dataId, dict):
         return dataId
+    elif hasattr(dataId, "mapping"):  # dafButler.dimensions.DataCoordinate
+        return {str(k): v for k, v in dataId.mapping.items()}
     elif hasattr(dataId, "items"):  # dafButler.dimensions.DataCoordinate
         return {str(k): v for k, v in dataId.items()}  # str() required due to full names
-    elif hasattr(dataId, "dataId"):  # dafButler.DimensionRecord
-        return {str(k): v for k, v in dataId.dataId.items()}
+    elif hasattr(dataId, "dataId"):  # dafButler.dimensions.DimensionRecord
```
mfisherlevine, line 485: Why did you add the `.dimensions` here? It's exported to the module level, so it wasn't wrong before.

python/lsst/summit/utils/nightReport.py
```diff
@@ -627,7 +627,7 @@ def plotPerObjectAirMass(
 
         objects = ensure_iterable(objects)
 
-        plt.figure(figsize=(10, 6))
+        plt.figure(figsize=(16, 12))
```
mfisherlevine, line 630: What is the purpose of changing the figure size here and below? I wonder if there's a better fix, as this will probably have other consequences. I suspect these defaults were picked quite carefully (and they're also not the same, and you've changed them to be the same).

tests/test_butlerUtils.py
```diff
@@ -64,6 +66,17 @@
...
 
+def custom_show(*args, **kwargs):
+    pass
+
+
+# Override the show function with our custom one
+# to avoid getting the next Warning:
+# UserWarning: FigureCanvasAgg is non-interactive, and thus cannot be shown
+# plt.show()
+plt.show = custom_show
```
mfisherlevine, line 77: As with the other review, I wonder if we can just remove the calls to `.show()` instead.

tests/test_tmaUtils.py
```diff
@@ -322,7 +322,7 @@ def test_endToEnd(self):
     @vcr.use_cassette()
     def test_noDataBehaviour(self):
         eventMaker = self.tmaEventMaker
-        noDataDayObs = 19500101  # do not use 19700101 - there is data for that day!
+        noDataDayObs = 19600101  # do not use 19700101 - there is data for that day!
```
mfisherlevine, line 325: Why did you change this?

config/quickLookIsr.py
kkeijuro: ok

python/lsst/summit/utils/nightReport.py
kkeijuro: it complains about figure size, I tried several bigger numbers and found those but maybe better numbers can be found (that feep the relation 10/6):
tests/test_nightReport.py::NightReportTestCase::test_plotPerObjectAirMass
  /sdf/home/c/cpio/development/summit_utils/python/lsst/summit/utils/nightReport.py:654: UserWarning: Tight layout not applied. The bottom and top margins cannot be made large enough to accommodate all axes decorations.
    plt.tight_layout()

tests/test_butlerUtils.py
kkeijuro: So remove the show inside the code? The show() is not part from the test but from the function in the library. I guess you mean returning the matplotlib Plot and then the uplayer call the show() from it, (but that is a big change)

python/lsst/summit/utils/nightReport.py
mfisherlevine: OK, thanks. It's not the most important plot to be honest, so let's leave it like this, and if it looks bad we can fix it up later, thanks.

tests/test_butlerUtils.py
mfisherlevine: I think that's an acceptable change - I don't think any of this library code should be calling `show()`, and so I think removing it is fine, and if things upstream break, that fixing them there is the right way to do things.

tests/test_tmaUtils.py
mfisherlevine: OK, that sounds good then, thanks!

PR #104, opened 2024-05-13
DM-44304: Add consdbClient.

ktlim:


python/lsst/summit/utils/consdbClient.py
```diff
@@ -0,0 +1,412 @@
...
+        variable ``LSST_CONSDB_PQ_URL`` (the location of the publish/query
+        service).
+
+    Notes
+    -----
+    This client is a thin layer over the publish/query Web service, which
+    avoids having a dependency on database drivers.
+
+    It enforces the return of query results as Astropy Tables.
+
```
mfisherlevine, line 76: We don't usually put newlines here

python/lsst/summit/utils/consdbClient.py
```diff
@@ -0,0 +1,412 @@
...
+        instrument: `str`
+            Name of the instrument (e.g. ``LATISS``).
+        obs_type: `str`
+            Name of the observation type (e.g. ``Exposure``).
+
+        Returns
+        -------
+        table_name: `str`
+            Name of the appropriate flexible metadata table.
+
```
mfisherlevine, line 168: Another here.

python/lsst/summit/utils/consdbClient.py
```diff
@@ -0,0 +1,412 @@
...
+    Does not do any quoting.  Mostly to remove a level of list-making.
+
+    Parameters
+    ----------
+    *args: `str`
+        Each parameter is a URL part.
+
+    Returns
+    -------
+    url: `str`
```
mfisherlevine, line 48: It might not really matter, but I think we generally do
```
paramater : `type`
```
 _i.e._ with a space before the colon.

python/lsst/summit/utils/consdbClient.py
```diff
@@ -0,0 +1,412 @@
...
+            Name of the instrument (e.g. ``LATISS``).
+        obs_type: `str`
+            Name of the observation type (e.g. ``Exposure``).
+
+        Returns
+        -------
+        key_info: `list` [ `list` [ `str` | `None` ] ]
+            List of string lists, one for each key. Each string list contains
+            the name of the key, a data type string, and a documentation
+            string.  A unit string and UCD string may also be present.
```
mfisherlevine, line 231: This makes me wonder if some simple dataclasses might simplify the interfaces. Perhaps we can add that later if it looks helpful though, but any time I see lists of lists with sub lists of variable length it gives me pause, as that can be quite clunky for any downstream code to use.

python/lsst/summit/utils/consdbClient.py
```diff
@@ -0,0 +1,412 @@
...
+
+        Returns
+        -------
+        response: `requests.Response`
+            HTTP response from the server, with 200 status for success.
+        """
+        if values:
+            values.update(kwargs)
+        else:
+            values = kwargs
```
mfisherlevine, line 302: I see how this works, but I'm not sure why you do it like this - I'm more just curious how you see this being used than anything else. Why allow an empty values dict and support kwargs?

Other than the explanation, the only comment would therefore be to add an `if not values: return` or something like that, as this opens up the possibility that there are no kwargs and also no values supplied (maybe with a warning to boot, too, actually).

python/lsst/summit/utils/consdbClient.py
```diff
@@ -0,0 +1,412 @@
...
```
mfisherlevine, line 351: Same comment here, maybe just check you're doing something before proceeding.

python/lsst/summit/utils/consdbClient.py
```diff
@@ -0,0 +1,412 @@
...
+        Notes
+        -----
+        This is a very general query interface because it is expected that
+        a wide variety of types of queries will be needed. If some types prove
+        to be common, syntactic sugar could be added to make the simpler.
+        """
+        url = _urljoin(self.url, "query")
+        data = {"query": query}
+        result = self._handle_post(url, data).json()
+        return Table(rows=result["data"], names=result["columns"])
```
mfisherlevine, line 387: I see that the obs_ids are getting cast to `str` on insertion. Does this bring them back to being ints on reading? If not, I think that would be worth adding, to make sure the type is preserved while round-tripping (especially on the primary key).

python/lsst/summit/utils/consdbClient.py
mfisherlevine: Having seen the `schema()` function below, maybe at least coercing these two things to always have the same shape by adding empty strings when the types or units are not there? (And in the `schema()` function, making that be the same length, as that doesn't look like it carries a UCD string ever?

python/lsst/summit/utils/consdbClient.py
```diff
@@ -0,0 +1,412 @@
...
+            contains the name of the key, a data type string, and a
+            documentation string.
+
+        Notes
+        -----
+        Fixed metadata data types may use the full database vocabulary,
+        unlike flexible metadata data types.
+        """
+        url = _urljoin(self.url, "schema", quote(instrument), quote(table))
+        return self._handle_get(url)
```
mfisherlevine, line 412: See comments on `get_flexible_metadata_keys()` about potentially having the same return shape.

tests/test_consdbClient.py
```diff
@@ -0,0 +1,140 @@
...
+from lsst.summit.utils import ConsDbClient
+
+
+@pytest.fixture
+def client():
+    return ConsDbClient("http://example.com/consdb")
+
+
+def test_table_name(client):
+    instrument = "latiss"
```
mfisherlevine, line 34: I know this isn't the real DB, but just noting here that I'm hoping that the instrument names in the real thing match the ones in the butler on case, _i.e._ they're like `LATISS`, `LSSTCam`, `LSSTComCam` etc. Or are they all lower-cased? I guess that wouldn't be the end of the world, though having them be the same would probably be preferable.

tests/test_consdbClient.py
```diff
@@ -0,0 +1,140 @@
...
+            responses.matchers.json_params_matcher({"key": "bar", "dtype": "int", "doc": "int key"}),
+        ],
+    )
+    responses.post(
+        "http://example.com/consdb/flex/bad_instrument/exposure/addkey",
+        status=404,
+        json={"message": "Unknown instrument", "value": "bad_instrument", "valid": ["latiss"]},
+    )
+    responses.post(
+        "http://example.com/consdb/flex/latiss/bas_obs_type/addkey",
```
mfisherlevine, line 73: Probably doesn't matter, but did you mean to write `bas_obs_type` or `bad_obs_type`?

tests/test_consdbClient.py
```diff
@@ -0,0 +1,140 @@
...
+        "http://example.com/consdb/flex/latiss/exposure/schema",
+        json=description,
+    )
+    instrument = "latiss"
+    obs_type = "exposure"
+    assert client.get_flexible_metadata_keys(instrument, obs_type) == description
+
+
+@responses.activate
+def test_get_flexible_metadata(client):
```
mfisherlevine, line 97: Great to see these types being checked, but I fear that because of how the mocking is done, this doesn't actually test the behaviour of the DB. I guess that's probably all that can be done without significant extra work, but just noting that if it's possible to get something a little more thorough up once we have the real thing that would be good (though I definitely understand there will be reluctance to using `vcr` for that, but perhaps there's something else we can do there).

tests/test_consdbClient.py
```diff
@@ -0,0 +1,140 @@
...
+    table = "misc_table"
+    assert client.schema(instrument, table) == description
+
+
+# TODO: more POST tests
+#    client.insert_flexible_metadata(
+#        instrument, obs_type, obs_id, values, allow_update
+#    )
+#    client.insert(instrument, table, obs_id, values, allow_update)
+#    client.query(query)
```
mfisherlevine, line 140: Uncomment or remove?

tests/test_consdbClient.py
```diff
@@ -0,0 +1,140 @@
...
+        json={"message": "Unknown instrument", "value": "bad_instrument", "valid": ["latiss"]},
+    )
+    responses.post(
+        "http://example.com/consdb/flex/latiss/bas_obs_type/addkey",
+        status=404,
+        json={"message": "Unknown observation type", "value": "bad_obs_type", "valid": ["exposure"]},
+    )
+
+    assert client.add_flexible_metadata_key(instrument, obs_type, "foo", "bool", "bool key")["key"] == "foo"
+    assert client.add_flexible_metadata_key(instrument, obs_type, "bar", "int", "int key")["key"] == "int"
```
mfisherlevine, line 79: Shouldn't that `== "int"` on the end be a `"bar"`?

tests/test_consdbClient.py
```diff
@@ -0,0 +1,140 @@
...
+        status=404,
+        json={"message": "Unknown instrument", "value": "bad_instrument", "valid": ["latiss"]},
+    )
+    responses.post(
+        "http://example.com/consdb/flex/latiss/bas_obs_type/addkey",
+        status=404,
+        json={"message": "Unknown observation type", "value": "bad_obs_type", "valid": ["exposure"]},
+    )
+
+    assert client.add_flexible_metadata_key(instrument, obs_type, "foo", "bool", "bool key")["key"] == "foo"
```
mfisherlevine, line 78: Does this actually run for you? This line raises a `TypeError: 'Response' object is not subscriptable` for me 🤔

tests/test_consdbClient.py
```diff
@@ -0,0 +1,140 @@
...
+    assert client.add_flexible_metadata_key(instrument, obs_type, "bar", "int", "int key")["key"] == "int"
+    client.add_flexible_metadata_key("bad_instrument", obs_type, "error", "int", "instrument error")
+    client.add_flexible_metadata_key(instrument, "bad_obs_type", "error", "int", "obs_type error")
```
mfisherlevine, line 81: Do you not want to catch these responses and check the messages or return codes?

python/lsst/summit/utils/consdbClient.py
```diff
@@ -0,0 +1,394 @@
...
+        Returns
+        -------
+        result: `Any`
+            Result of decoding the Web service result content as JSON.
+        """
+        logger.debug(f"GET {url}")
+        try:
+            response = self.session.get(url, params=query)
+        except requests.exceptions.RequestException as e:
+            raise e
```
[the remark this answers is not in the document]
ktlim: I think it was for possibly retrying, but it's not doing anything now.

python/lsst/summit/utils/consdbClient.py
ktlim: Added a dataclass for flexible metadata, but `schema()` is native Postgres so can never have units or UCDs and has a larger dtype vocabulary, as mentioned in the Note, so I don't want it to look too similar.

python/lsst/summit/utils/consdbClient.py
ktlim: I thought this was convenient and similar to Butler data ids, where you can specify a dict or kwargs, with the latter overriding the former.

python/lsst/summit/utils/consdbClient.py
ktlim: They're cast to `str` to put in the URL, but they're handled as `int` otherwise, including here.

tests/test_consdbClient.py
ktlim: Since they're used in the database without quoting, they can actually be any case.

tests/test_consdbClient.py
ktlim: Indeed, these tests don't actually test the server behavior, just the request generation and response formatting.  The list of `responses` is I think effectively a `vcr` cassette, isn't it?

tests/test_consdbClient.py
ktlim: TODOs (future tests to be added, in this case) are allowed, I think.

tests/test_consdbClient.py
ktlim: Fixed.

tests/test_consdbClient.py
ktlim: Fixed.

PR #107, opened 2024-05-22
DM-44446: Add pending type annotations to summit_*

kkeijuro:
Add pending type annotations to summit_*


python/lsst/summit/utils/bestEffort.py
```diff
@@ -115,7 +115,7 @@ def __init__(
         self.defaultExtraIsrOptions = defaultExtraIsrOptions
 
         self._cache: dict = {}
-        self._cacheIsForDetector = None
+        self._cacheIsForDetector: str | None = None
```
mfisherlevine, line 118: detector is an `int` not a `str`.

python/lsst/summit/utils/bestEffort.py
```diff
@@ -263,7 +263,7 @@ def getExposure(
 
         if self._cacheIsForDetector != dataId["detector"]:
             self.clearCache()
-            self._cacheIsForDetector = dataId["detector"]
+            self._cacheIsForDetector = str(dataId["detector"])
```
mfisherlevine, line 266: Please leave this as an int.

python/lsst/summit/utils/butlerUtils.py
```diff
@@ -385,19 +385,19 @@ def getExpIdFromDayObsSeqNum(butler: dafButler.Butler, dataId: dict) -> dict:
     return {"exposure": expRecord.id}
 
 
-def updateDataIdOrDataCord(dataId: dict, **updateKwargs: Any) -> dict:
+def updateDataIdOrDataCord(dataId: dafButler.DataId, **updateKwargs: Any) -> Mapping[str, Any]:
```
mfisherlevine, line 388: `dataId` here can definitely also be a dict, and usually is, I think. Please make it be allowed to be both.

python/lsst/summit/utils/butlerUtils.py
```diff
@@ -411,19 +411,19 @@ def updateDataIdOrDataCord(dataId: dict, **updateKwargs: Any) -> dict:
     return newId
 
 
-def fillDataId(butler: dafButler.Butler, dataId: dict) -> dict:
+def fillDataId(butler: dafButler.direct_butler.DirectButler, dataId: dafButler.DataId) -> Mapping[str, Any]:
```
mfisherlevine, line 414: I'm surprised that the `DirectButler` part is necessary to be honest. The `dataId` also need to be able to take a dict though, that's the main thing.

python/lsst/summit/utils/butlerUtils.py
```diff
@@ -411,19 +411,19 @@ def updateDataIdOrDataCord(dataId: dict, **updateKwargs: Any) -> dict:
...
     """Given a dataId, fill it with values for all available dimensions.
 
     Parameters
     ----------
     butler : `lsst.daf.butler.Butler`
         The butler.
-    dataId : `dict`
+    dataId : `dafButler.DataId`
```
mfisherlevine, line 421: dict here too.

python/lsst/summit/utils/butlerUtils.py
```diff
@@ -521,14 +523,14 @@ def getExpRecordFromDataId(butler: dafButler.Butler, dataId: dict) -> dafButler.
...
 
-    expRecords = set(expRecords)
-    if not expRecords:
+    filteredExpRecords = set(expRecords)
+    if not filteredExpRecords:
         raise LookupError(f"No exposure records found for {dataId}")
-    assert len(expRecords) == 1, f"Found {len(expRecords)} exposure records for {dataId}"
-    return expRecords.pop()
+    assert len(filteredExpRecords) == 1, f"Found {len(filteredExpRecords)} exposure records for {dataId}"
+    return filteredExpRecords.pop()
```
mfisherlevine, line 530: Was all that necessary just to make mypy happy because it would change the type to being a `set`? If so, that's fine, but it really does make me unhappy that that sort of thing is necessary. I wonder if there's a better way to make it happy...

python/lsst/summit/utils/butlerUtils.py
```diff
@@ -559,17 +561,17 @@ def getDayObsSeqNumFromExposureId(butler: dafButler.Butler, dataId: dict) -> dic
...
     )
-    expRecords = set(expRecords)
-    if not expRecords:
+    filteredExpRecords = set(expRecords)
+    if not filteredExpRecords:
         raise LookupError(f"No exposure records found for {dataId}")
-    assert len(expRecords) == 1, f"Found {len(expRecords)} exposure records for {dataId}"
-    record = expRecords.pop()
+    assert len(filteredExpRecords) == 1, f"Found {len(filteredExpRecords)} exposure records for {dataId}"
+    record = filteredExpRecords.pop()
```
mfisherlevine, line 568: Same here.

python/lsst/summit/utils/consdbClient.py
```diff
@@ -485,4 +485,4 @@ def schema(self, instrument: str, table: str) -> dict[str, tuple[str, str]]:
         """
         url = _urljoin(self.url, "schema", quote(instrument), quote(table))
         result = self._handle_get(url)
-        return {key: tuple(value) for key, value in result.items()}
+        return {key: (str(value[0]), str(value[1])) for key, value in result.items()}
```
mfisherlevine, line 488: I think you need to check with @ktlim that these are actually strings and that cast is OK.

python/lsst/summit/utils/imageExaminer.py
```diff
@@ -451,7 +456,7 @@ def plotRadialAverage(self, ax: matplotlib.axes.Axes | None = None) -> None:
 
         ax.plot(distances, values, "x", label="Radial average")
         if not fitFailed:
-            fitline = gauss(distances, *pars)
+            fitline = gauss(np.array(distances), *pars)
```
mfisherlevine, line 459: `distances` is a list of floats, and that obviously works, because the code runs fine, so I think in cases like that it's better to add to the annotations rather than doing an unnecessary cast, so I'd just add that lists are OK for the `gauss` function.

python/lsst/summit/utils/imageExaminer.py
```diff
@@ -584,8 +589,12 @@ def plotFullExp(self, ax: matplotlib.axes.Axes | None = None) -> None:
...
 
         imData = quickSmooth(self.exp.image.array, 2.5)
-        vmin = np.percentile(imData, 10)
-        vmax = np.percentile(imData, 99.9)
+        # np.percentile returns a float if imData is a 1D array,
+        # but it can return a numpy array if imData is a 2D array
+        # or higher. If imData is a 2D array, vmin and vmax will
+        # be arrays, which is not what LogNorm expects.
+        vmin = np.percentile(imData, 10).item()
+        vmax = np.percentile(imData, 99.9).item()
```
mfisherlevine, line 597: but `imData` here _is_ 2d, and it's giving a scalar already, so that's not necessary for the code to work - it will only give you an array if you specify `axis=0` (or `1`), so I don't love this. Is it really giving errors? I didn't know about calling `.item()` so I find that a little confusing and would prefer to avoid that if it's possible.

python/lsst/summit/utils/nightReport.py
```diff
@@ -199,8 +199,8 @@ def getExpRecordDictForDayObs(self, dayObs: int) -> dict:
         expRecords = self.butler.registry.queryDimensionRecords(
             "exposure", where="exposure.day_obs=dayObs", bind={"dayObs": dayObs}, datasets="raw"
         )
-        expRecords = list(expRecords)
-        records = {e.seq_num: e.toDict() for e in expRecords}  # not guaranteed to be in order
+        listExpRecords = list(expRecords)
+        records = {e.seq_num: e.toDict() for e in listExpRecords}  # not guaranteed to be in order
```
mfisherlevine, line 203: Again, I kind of hate this type of workaround. Is there not a more elegant way to tell mypy we're changing types than to make new variables?

python/lsst/summit/utils/nightReport.py
```diff
@@ -762,7 +762,7 @@ def makeAltAzCoveragePlot(
         ax.set_title("Axial coverage - azimuth (theta, deg) vs zenith angle (r, deg)", size=20)
 
         for h in lgnd.legend_handles:
-            if h is None:
+            if h is None or not isinstance(h, matplotlib.lines.Line2D):
```
mfisherlevine, line 765: Kind of gross, but I guess this is because matplotlib hasn't properly annotated the return type above? Or it has, and it can return other things?

python/lsst/summit/utils/peekExposure.py
```diff
@@ -331,7 +339,7 @@ def run(self, exposure: afwImage.Exposure, binSize: int | None = None) -> pipeBa
             exposure.setMaskedImage(binned)
 
         if self.config.doInstallPsf:
-            self.installPsf.run(exposure=exposure)
+            self.installPsf.runa(exposure=exposure)
```
mfisherlevine, line 342: What is this? I'm worried on a few counts here - firstly, that looks like a random typo that will break the code got committed, but also, @jmeyers314 does `PeekExposureTask` not have any tests which check that it runs? Or is it just that it doesn't run with that `InstallGaussianPsfTask` set to run?

python/lsst/summit/utils/bestEffort.py
kkeijuro: Tried to find the variable type and finally decided to put as str...

mfisherlevine: You can always ask - always better than guessing 🙂

python/lsst/summit/utils/bestEffort.py
kkeijuro: Sure

python/lsst/summit/utils/butlerUtils.py
kkeijuro: DataId definition include also a Mapping, dict is also a Mapping and in the use on this function the Mapping is enough, but I can change it to dict (more general)

python/lsst/summit/utils/butlerUtils.py
kkeijuro: I will try to redeclare the type....

kkeijuro: But not working...

python/lsst/summit/utils/imageExaminer.py
kkeijuro: list[float] - float is not accepted... And I dont know why its working with a list of float, I can remove the conversion and add a type: ignore, sincerelly is the only idea I have

python/lsst/summit/utils/imageExaminer.py
kkeijuro: Problem is in:
LogNorm(vmin=vmin, vmax=vmax)
Which only accepts vmin and vmax to be a float,

python/lsst/summit/utils/nightReport.py
kkeijuro: I've tried several other things and didnt work, but I will as in #sofware-dev

python/lsst/summit/utils/nightReport.py
kkeijuro: It can return other things...

python/lsst/summit/utils/peekExposure.py
kkeijuro: I ask in #software-dev and paul price gave me this solutions...

kkeijuro: of course runa is a mistake... sorry!

python/lsst/summit/utils/butlerUtils.py
kkeijuro: Now I check we are able to control this error using the allow_redefinition = True (I had it to false)

python/lsst/summit/utils/peekExposure.py
jmeyers314: No tests for `PeekExposureTask` sadly, unless you count the non-CI'd `assessPeekExposure` script: https://github.com/lsst-sitcom/summit_utils/blob/cf0e52989bfda98354aa57ec6a4e5b3a4ba50e41/python/lsst/summit/utils/assessPeekExposure.py#L47

python/lsst/summit/utils/bestEffort.py
kkeijuro: OK, now need an assert or a typing: ignore to remove mypy error

PR #109, opened 2024-06-03
DM-44656: Enable PeekExposureTask for LSSTComCam

jmeyers314:


python/lsst/summit/utils/peekExposure.py
```diff
@@ -968,10 +970,16 @@ def runPeek(
...
-                mode = "photo"
+                try:
+                    isDisp = isDispersedExp(exposure)
+                except Exception as e:
+                    self.log.warning(f"isDispersedExp failed: {e}.  Assuming not dispersed")
+                    isDisp = False
+                if isDisp:
+                    mode = "spec"
+                else:
+                    mode = "photo"
```
mfisherlevine: There's two things I don't love here: exceptions as control flow (not so bad given this won't be in a tight loop), and the fact this will routinely spew really quite scary looking warnings (made somewhat worse still because it includes the whole "{e}" in the message) during normal/expected operation (which is really bad behaviour IMO). I think it shouldn't be too hard to avoid that though. I've just pushed a commit (untested) that I think should give you what you want though. Could you test that it does what you want? If it does, then that's a 👍 from me (obviously).

jmeyers314: Thanks, that looks good to me.  I've reordered some of the if/else logic and removed the unnecessary `isDisp` variable, but followed your same logic.

python/lsst/summit/utils/peekExposure.py
```diff
@@ -968,10 +970,13 @@ def runPeek(
...
-            elif isDispersedExp(exposure):
-                mode = "spec"
             else:
-                mode = "photo"
+                if exposure.getInfo().getVisitInfo().instrumentLabel == "LATISS":
+                    # only LATISS images *can* be dispersed, and isDispersedExp
+                    # only works cleanly for LATISS
+                    mode = "spec" if isDispersedExp(exposure) else "photo"
+                else:
+                    mode = "photo"
```
mfisherlevine, line 979: FWIW, and now we're probably just in the land of personal preference, but I'd write that as:
```
mode = "photo"
if exposure.getInfo().getVisitInfo().instrumentLabel == "LATISS":
    # only LATISS images *can* be dispersed, and isDispersedExp
    # only works cleanly for LATISS
    mode = "spec" if isDispersedExp(exposure)
```
because otherwise you end up with a variable that's defined inside if/else blocks which could end up, if the logic changed in the future, undefined, and this way you'll always have the fallback. Definitely in the noise though 🙂

mfisherlevine: Oh, except that's a supplied arg... anyway, whatever, all is good, this looks great 🙂 👍

PR #115, opened 2024-06-22
DM-44917: Add detector utilities

mfisherlevine:


python/lsst/summit/utils/utils.py
```diff
@@ -1203,3 +1207,56 @@ def computeCcdExposureId(instrument: str, exposureId: int, detector: int) -> int
...
+
+
+def getCameraFromInstrumentName(instrumentName: str) -> lsst.afw.cameraGeom.Camera:
+    """Get the camera object given the instrument name (case insenstive).
+
+    Parameters
+    ----------
+    instrumentName : `str`
+        The name of the instrument, e.g. "LATISS" or "LSSTCam". Case
+        insenstive.
```
TallJimbo, line 1236: So these are the same names as used in the butler data IDs, right?  You're just being more permissive about case?

TallJimbo (approved):

python/lsst/summit/utils/utils.py
mfisherlevine: Yes, exactly. ConsDB is all lower case, so it's now quite useful to do so.

PR #68, opened 2023-11-21
DM-34340: Adapt to daf_butler API changes.

TallJimbo:

PR #116, opened 2024-07-12
DM-45226: Change np.float_ to np.float64

timj:
np.float_ has been removed in numpy 2.


python/lsst/summit/utils/imageExaminer.py
```diff
@@ -513,6 +513,10 @@ def plotSurface(self, ax: Axes3D | None = None, useColor: bool = True) -> None:
             fig, ax = plt.subplots(subplot_kw={"projection": "3d"}, figsize=(10, 10))
             plotDirect = True
 
+        # Allow mypy to know that this is a single Axes3D and not list[Axes]
+        # or Any or None. plt.subplots is annotated to return Any.
+        assert isinstance(ax, Axes3D)
```
timj, line 518: This assert breaks `test_quickLook.py`. Mypy does have a point that `plt.subplots` could return many things and it doesn't know.

I think the mypy breakage might be from the new 3.9.1 matplotlib which noted they had enhanced their type annotations. There is also a breakage in

```
python/lsst/summit/utils/m1m3/plots/inertia_compensation_system.py:279: error: Incompatible return value type (got "Rectangle", expected "Polygon")  [return-value]
```
which is confusing because the documentation says that `Polygon` is the right thing.

mfisherlevine: Thanks Tim - I've also been confused by these things for a while now (and am very new to mypy, so never know if it's me or matplotlib, or how things are configured). I'm off on vacation today and will be back online on Monday, so feel free to do whatever you see as best in my absence - Colin and I are being quite sanguine about the fact that the next couple of weeks is the right time to break things in order to make progress (other breaking changes in both summit_utils and RA are coming up), so if that does happen we're understanding that it's not the end of the world right now.

mfisherlevine: (by which I mean that I can't really review this properly until I'm back, but if you end up blocked by this and wanting to merge it, just go for it, and we can deal with any problems it might cause, but it seems like it should improve things, rather than the opposite).

mfisherlevine (approved):
I think this is all fine, and I think that Jim's PR for DM-34340 fixes the bit where we've been talking, so once he merges that and this is rebased I think this is good to go.

python/lsst/summit/utils/imageExaminer.py
timj: @TallJimbo fixed this on #68.

PR #120, opened 2024-09-04
Redact tokens/passwords from URLs in ConsDbClient HTTPError reports

Vebop:
## Added `clean_url()` to `requests.Session` hooks for responses. 
Added test to check initialization of ConsDbClient and to test that the url is sanitized according to our terms.

### DIFF
--- a/python/lsst/summit/utils/consdbClient.py
+++ b/python/lsst/summit/utils/consdbClient.py
@@ -23,7 +23,7 @@ import logging
 import os
 from dataclasses import dataclass
 from typing import Any
-from urllib.parse import quote
+from urllib.parse import quote, urlparse
 
 import requests
 from astropy.table import Table
@@ -77,6 +77,24 @@ def _check_status(r: requests.Response) -> None:
         raise e
 
 
+def clean_url(resp: requests.Response, *args, **kwargs) -> requests.Response:
+    """Parse url from response and remove netloc portion.
+
+    Set new url in response and return response
+
+    Parameters
+    ----------
+    resp : `requests.Response`
+        The response that could contain a URL with tokens
+    """
+    url = urlparse(resp.url)
+    short_user = f"{url.username[:2]}***" if url.username is not None else ""
+    short_pass = f":{url.password[:2]}***" if url.password is not None else ""
+    netloc = f"{short_user}{short_pass}@{url.hostname}"
+    resp.url = url._replace(netloc=netloc).geturl()
+    return resp
+
+
 @dataclass
 class FlexibleMetadataInfo:
     """Description of a flexible metadata value.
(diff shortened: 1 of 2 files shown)


bbrondel:
I'd think you would want to redact the entire password, including the first two letters, and for that matter avoid revealing anything about the length of the password.

mfisherlevine:
Also, I'm not usually a stickler for asking people to squash commits, but given the middle commits include adding and removing a whole file which should never have been committed, and that this whole ticket feels fine as a single commit, I'd probably just squash the whole thing down to one or two commits.

python/lsst/summit/utils/consdbClient.py
```diff
@@ -77,6 +77,24 @@ def _check_status(r: requests.Response) -> None:
...
+
+    Set new url in response and return response
+
+    Parameters
+    ----------
+    resp : `requests.Response`
+        The response that could contain a URL with tokens
+    """
+    url = urlparse(resp.url)
+    short_user = url.username[:2] if url.username is not None else "**"
```
ktlim, line 91: I think it's a bit more useful to have
```
short_user = f"{url.username[:2]}***" if url.username is not None else ""
short_pass = f":{url.password[:2]}***" if url.password is not None else ""
netloc = f"{short_user}{short_pass}@{url.hostname}"
```
That way, empty usernames and passwords are properly reflected and the password component is only included if it is actually present.

tests/test_consdbClient.py
```diff
@@ -28,6 +28,9 @@
 
 @pytest.fixture
 def client():
+    """Initilaize client with a fake url
```
ktlim, line 31:
```suggestion
    """Initialize client with a fake url
```

tests/test_consdbClient.py
```diff
@@ -161,6 +164,38 @@ def test_schema(client):
...
+        )
+
+    url = error.value.args[0].split()[-1]
+    sanitized = (
+        "https://us***:v9***@usdf-fake.slackers.stanford.edu/consdb/flex/bad_instrument/exposure/addkey"
+    )
+
+    assert "v987wefVMPz" not in url
+    assert url == sanitized
+
```
ktlim, line 192: If you really want coverage, you could also test the username/no password case, a password/no username case (although I don't know if that is actually legal), and a single-character username case (if someone uses 1 or 2 character passwords, that's their own problem...).

ktlim (approved):
General approach looks great.

Vebop:
Updated the `short_user` and `short_pass` to be empty if none are provided; 
Checked multiple test cases in a parametrized unit test; 
Reorganized commits -- but don't we use a squash commit :)

PR #121, opened 2024-09-26
DM-46086: Make EFD query raising on no schema entry an optional parameter

mfisherlevine:


python/lsst/summit/utils/efdUtils.py
```diff
@@ -308,7 +321,10 @@ async def _getEfdData(
     availableTopics = await client.get_topics()
 
     if topic not in availableTopics:
-        raise ValueError(f"Topic {topic} not in EFD schema")
+        if raiseIfTopicNotInSchema:
+            raise ValueError(f"Topic {topic} not in EFD schema")
+        else:
+            return pd.DataFrame()
```
isullivan, line 327: Do you need to ensure that the empty dataframe contains any specific columns, or has an index?

isullivan: Also, I suggest adding a warning (could be at `VERBOSE` level) stating that the topic was not in the schema and empty data is being returned.

isullivan (approved):
One question about the empty dataframe. Otherwise, I am not sure where it should go but it would be helpful to have a comment somewhere (maybe the docstring in efdUtils?) about why you might want to not raise if the topic is not in the schema.

python/lsst/summit/utils/efdUtils.py
mfisherlevine: Ha, if only data in the EFD was that predictable! But yeah, columns don't matter, the main thing is that it doesn't raise - downstream code that's setting the "don't raise if the topic isn't there" is then responsible for dealing with the lack of data.

mfisherlevine: By `VERBOSE` I assume you mean `DEBUG`, in which case sure 👍

mfisherlevine:
> One question about the empty dataframe. Otherwise, I am not sure where it should go but it would be helpful to have a comment somewhere (maybe the docstring in efdUtils?) about why you might want to not raise if the topic is not in the schema.

Good point. I think much the most called function here is `getEfdData`, so I'll add it to the docstring there, as I think that's the only place it really fits. The answer though reveals why it's not so important though: `raiseIfTopicNotInSchema` should be set to `False` when running on the summit or in utility code for topics which might have no data taken in within the last <data_retention_period> (nominally 30 days). Once a topic is in the schema at USDF it will always be there, and thus users there never need worry about this, always using `False` is fine, but at the summit things are a little less predictable, so something missing from the schema doesn't _necessarily_ mean a typo (but it still will 99% of the time).

PR #122, opened 2024-10-03
DM-46527: Add client methods for waiting for data to land in ConsDB

mfisherlevine:


python/lsst/summit/utils/consdbClient.py
```diff
@@ -607,6 +608,130 @@ def query(self, query: str) -> Table:
...
+        # we know the row now exists but the required item may not be there yet
+        sleep_duration = 1 / poll_frequency_hz
+        t0 = time.time()
+        while time.time() - t0 < timeout:
+            result = self.query(query)
+            if len(result) > 1:
+                raise ValueError(f"Query {query} returned more than one row")
+            assert len(result) == 1, "Somehow no rows came back, which should be impossible"
+            row = result[0]
+            if item not in row.columns:
```
yalsayyad, line 723: I wasn't sure what you meant by "item" (in the doc string and method name) until I got to this row. In a relational database, the intersection of a column and a row is sometimes called a "field" and the contents are the field value. Sometimes a column itself is called a "field" (or "attribute").  You could probably go with "value", "field value", "field."

From what I can tell the input parameter is "column"/"attribute" flavored and the return is a "field"/"value"/"fieldValue" flavored.

yalsayyad (approved):

PR #124, opened 2024-10-12
DM-45796: Stop night report crashing when nothing observed

mfisherlevine:


fritzm (approved):
LGTM

python/lsst/summit/utils/nightReport.py
```diff
@@ -736,8 +736,13 @@ def makeAltAzCoveragePlot(
             objects = self.stars
         objects = ensure_iterable(objects)
 
+        if not objects:  # there's genuinely nothing now
+            self.log.info("No objects to plot")
+            return None
+
```
laurenam, line 742: While I think it will just return the empty plot in that case, but do you want to do the same check and bail with error message in `plotPerObjectAirMass()`?  If just for consistency...

laurenam (approved):

python/lsst/summit/utils/nightReport.py
mfisherlevine: Agreed.

PR #125, opened 2024-10-28
DM-47189: Write default butler factory function for all instruments

mfisherlevine:


python/lsst/summit/utils/butlerUtils.py
```diff
@@ -150,6 +158,75 @@ def makeDefaultLatissButler(
...
+    try:
+        butler = dafButler.Butler.from_config(
+            repo, collections=collections, writeable=writeable, instrument=instrument
+        )
+    except (FileNotFoundError, RuntimeError):
+        # Depending on the value of DAF_BUTLER_REPOSITORY_INDEX and whether
+        # it is present and blank, or just not set, both these exception
+        # types can be raised, see tests/test_butlerUtils.py:ButlerInitTestCase
+        # for details and tests which confirm these have not changed
+        raise FileNotFoundError  # unify exception type
```
isullivan, line 226: Do you want to include the original error message?

python/lsst/summit/utils/butlerUtils.py
```diff
@@ -150,6 +158,75 @@ def makeDefaultLatissButler(
...
+        Whether to make a writable butler.
+    embargo : `bool`, optional
+        Use the embargo repo instead of the main one. Needed to access
+        embargoed data if not at a summit-like location.
+
+    Returns
+    -------
+    butler : `lsst.daf.butler.Butler`
+        The butler.
+    """
```
isullivan, line 185: You could add a `raises` section for the `FileNotFoundError` handling.

isullivan (approved):
A couple minor suggestions.

python/lsst/summit/utils/butlerUtils.py
mfisherlevine: This should be _incredibly_ rare tbh. I almost removed this who `try` really - this was put in in the very very early days before we reliably had the `$DAF_BUTLER_REPOSITORY_INDEX` set at all our sites, and sometimes it was set but pointed to a file that was missing, etc etc. I suspect this will simply never happen now, so I think this is OK.

python/lsst/summit/utils/butlerUtils.py
mfisherlevine: I will add the `raises` section for good measure though.

PR #126, opened 2024-10-29
DM-47269:  Change fail behavior of getSite() and fix build breakage

mfisherlevine:


python/lsst/summit/utils/utils.py
```diff
@@ -599,12 +599,9 @@ def getSite() -> str:
     location : `str`
         One of:
         ['tucson', 'summit', 'base', 'staff-rsp', 'rubin-devl', 'jenkins',
-         'usdf-k8s']
+         'usdf-k8s'].
```
isullivan, line 602: Really minor, but I don't think you want the added period.

python/lsst/summit/utils/utils.py
```diff
@@ -647,7 +644,7 @@ def getSite() -> str:
         return "usdf-k8s"
 
     # we have failed
-    raise ValueError("Location could not be determined")
+    return "UNKOWN"
```
isullivan, line 647: Note that butlerUtils and efdUtils each contain a `try..except` block that checks for `ValueError`. Suggest modifying those to remove the `try..except` or just ticketing the change.

isullivan (approved):

python/lsst/summit/utils/utils.py
mfisherlevine: Thanks. Changed one, and ticketed the one that appears to be fine and I'm too scared to touch right now!

PR #127, opened 2024-11-08
DM-46784: Add function for getting ecliptic coordinates

mfisherlevine:


TallJimbo (approved):
I know you probably can't do it now, but if you're reimplementing something from astropy for performance reasons, it's both particularly important and particularly easy to add a unit test.

mfisherlevine:
> I know you probably can't do it now, but if you're reimplementing something from astropy for performance reasons, it's both particularly important and particularly easy to add a unit test.

Good shout - done 🙂

PR #128, opened 2024-11-08
DM-47477: Add M1M3 hardpoint measurements to TMA SLEW plots

craiglagegit:


craiglagegit:
Hi @mfisherlevine .  So this is all working.  Attached are two examples.  One is a recent plot moving at 2%, and one is from January moving at 70%.  I fixed all of the formatting errors but one.  There is still a mypy error that I cannot figure out.
Slew_Plot_20240102_1308
Slew_Plot_20241107_90

PR #137, opened 2025-01-31
DM-45545: Conform code with new flake8 and E226

timj:


isullivan (approved):
These changes appear consistent with E226

PR #143, opened 2025-04-14
DM-50161 Add token as an optional constructor parameter in ConsDbClient.

bbrondel:
Though it's possible to use ConsDbClient from outside of RSP by adding a token to the URL, this patch allows the user to pass the token through the HTTP header as an alternative.


python/lsst/summit/utils/consdbClient.py
```diff
@@ -135,6 +135,9 @@ class ConsDbClient:
         variable ``LSST_CONSDB_PQ_URL`` (the location of the publish/query
         service).
```
ktlim, line 137: No blank line here, I believe.

ktlim (approved):

PR #146, opened 2025-04-19
DM-50347: Add pixInsight stretch

mfisherlevine:


python/lsst/summit/utils/plotting.py
```diff
@@ -233,16 +220,21 @@ def plot(
                 norm = vis.ImageNormalize(imageData, interval=interval, stretch=vis.LinearStretch())
             case "sqrt":
                 norm = vis.ImageNormalize(imageData, interval=interval, stretch=vis.SqrtStretch())
+            case "pixInsight":
+                imageData = stretchDataPixInsight(imageData)
+                norm = None
```
isullivan, line 225: Since mypy complains about this line, you could probably do:
`norm = vis.ImageNormalize(imageData, stretch=vis.LinearStretch())`

python/lsst/summit/utils/plotting.py
```diff
@@ -322,3 +314,191 @@ def plot(
...
+    image : `np.ndarray`
+        Image after applying the MTF, with NaNs replaced by fallback values.
+    """
+    M = np.full(image.shape, midtonesBalance)
+    maskHalf = image == 0.5
+    maskZero = image == 0
+    maskOne = image == 1
+    fallback = (maskHalf * 0.5) * (1 - maskZero) + maskOne
+    result = (M - 1) * image / ((2 * M - 1) * image - M)
+    nanMask = np.isnan(result)
```
isullivan, line 341: Perhaps use `~np.isfinite(result)` to catch inf values too.

isullivan (approved):
A couple minor suggestions, but OK however you address them.

python/lsst/summit/utils/plotting.py
mfisherlevine: Oh nice, I actually didn't run a final mypy after that commit, thanks!

python/lsst/summit/utils/plotting.py
mfisherlevine: Good shout, thanks 🙏

PR #150, opened 2025-05-26
DM-51096: Use toy pointing model to calculate mount error

craiglagegit:
There have been several recent exposures where the images were obviously streaked, but the RubinTV Mount Torque plots reported no issue.  This is because the difference between the demand position and the actual position was small because the demand position was not doing what it should.  Examples are 2025051900417, where the mount was not tracking, and 2025052000321, where the rotator was still moving rapidly during the exposure.  Plots and images attached.  In both cases the Mount motion image degradation reported a small value even though the images were unacceptable.  The idea is to install a simplified MTPtg model in the mount analysis code in summit_utils to try and flag these.  This code seems to do that.

### DIFF
--- a/python/lsst/summit/utils/simonyi/mountAnalysis.py
+++ b/python/lsst/summit/utils/simonyi/mountAnalysis.py
@@ -42,6 +42,7 @@ from lsst.summit.utils.utils import dayObsIntToString
 from .mountData import getAzElRotDataForExposure
 
 if TYPE_CHECKING:
+    from astropy.time import Time
     from lsst_efd_client import EfdClient
     from matplotlib.figure import Figure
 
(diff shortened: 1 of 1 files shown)


craiglagegit:
I made the above changes that we discussed.  Attached are plots of three different expIds, one good one and two of the trailed images from last week that weren't caught.  A simple linear model of the Az/El/Rot changes seems to work fine.  The way it works is that the standard method is to calculate the RMS(actualPosition-demandPosition).  Here we first calculate RMS(actualPosition-modelPosition)  and if that RMS is less than some value (here set to 1.0 arcsec for Az and El and 5.0 arcsec for Rot), we use the standard RMS.  If it is greater than that value , then we replace the (actualPosition-demandPosition) errors with the (actualPosition-modelPosition) errors.  Here is a PR, but maybe more changes are needed.
New_Mount_Plot_2025051900417
New_Mount_Plot_2025052000245
New_Mount_Plot_2025052000321

mfisherlevine (approved):

PR #151, opened 2025-06-04
DM-51257: Add hexapod motion to plots and image residual calculation

craiglagegit:
Here's an example of the new proposed plot.
Mount_Plot_Hex_2025052500498


craiglagegit:
Ha!  All checks passed on the first try!!

craiglagegit:
Hi @mfisherlevine .  I think this is ready to merge.  Please review and let me know what you think.  Thanks!

PR #159, opened 2025-07-31
OSW-22 hotfix: Fix type hinting.

bbrondel:


mfisherlevine:
Commits need squashing, other than that, it looks good.

PR #161, opened 2025-08-07
DM-52072: Switch to using reported hexapod errors

craiglagegit:
Adding a better way to calculate the hexapod RMS values, trying to eliminate all of the yellow values.


craiglagegit:
I compared the new method to the old for one night (20250718), with a lot of failures.  The RMS on the failure is smaller, but still large enough to catch the failures.  The RMS on the good ones are also smaller, which should eliminate most of the yellw ones.
DM-52072_New_vs_Old_20250718

PR #167, opened 2025-11-13
DM-49364: Switch LATISS/bestEffortIsr to IsrTaskLSST

mfisherlevine:


config/quickLookIsr.py
```diff
@@ -1,6 +1,7 @@
-config.doWrite = False  # this task writes separately, no need for this
 config.doSaturation = True  # saturation very important for roundness measurement in qfm
-config.doSaturationInterpolation = True
-config.overscan.fitType = "MEDIAN_PER_ROW"
-config.overscan.doParallelOverscan = True
 config.brighterFatterMaxIter = 2  # Uncomment this to remove test warning
+config.doDeferredCharge = False  # no calib for this yet
+config.doBootstrap = False
+config.doApplyGains = False
```
czwa: Unless you expect to never have a PTC, I would strongly suggest this be set to `doApplyGains=True`.  The new ISR defaults largely assume that everything has had gains applied so units are in electrons.  We also should have CTI calibrations in the next set, but that correction is far less likely to have visual impact on the images than `doApplyGains` (which should help remove amp offsets).

czwa (approved):
I'm not going to make the change request mandatory, but unless there's a compelling reason not to `doApplyGains=True`, I think that will improve things.

PR #185, opened 2026-02-13
DM-53212: Add functions for getting full ccdVisit table and wide quicklook table

mfisherlevine:


python/lsst/summit/utils/consdbClient.py
```diff
@@ -694,3 +695,137 @@ def schema(
...
+    v = client.query(f"SELECT * FROM cdb_LSSTCam.visit1 WHERE day_obs = {dayObs}")
+
+    wide: Table = join(
+        vq,
+        v,
+        keys="visit_id",
+        join_type="inner",
+        table_names=("vq", "v"),
+        uniq_col_name="{col_name}_{table_name}",  # only duplicates get suffixed
+    )
```
TallJimbo: I'm a little surprised you're doing this join with astropy instead of SQL.

TallJimbo (approved):

python/lsst/summit/utils/consdbClient.py
bbrondel: I really think it makes more sense to do this join with SQL.

mfisherlevine: Looks like I can replace this with
```
vqCols = set(client.query("SELECT * FROM cdb_LSSTCam.visit1_quicklook LIMIT 0").colnames)
vCols = set(client.query("SELECT * FROM cdb_LSSTCam.visit1 LIMIT 0").colnames)

vOnlyCols = vCols - vqCols  # exclude visit_id and all duplicates

selectClauses = ["vq.*"] + [f"v.{col}" for col in sorted(vOnlyCols)]

query = f"""
    SELECT {', '.join(selectClauses)}
    FROM cdb_LSSTCam.visit1_quicklook vq
    INNER JOIN cdb_LSSTCam.visit1 v USING (visit_id)
    WHERE vq.day_obs = {dayObs}
"""
return client.query(query)
```
Does that look reasonable to you, or am I still missing tricks?

mfisherlevine: And was it just this function, or did you think the other one should have more things moved to SQL too?

TallJimbo: That looks good to me, and think it was only this function.

PR #186, opened 2026-02-25
Fix FootprintSet STDEV detection failure in runSourceDetection

estevesjh:
## Summary
  - Add uniform [0, 1) random dither per stamp before median coadd to break integer
  quantization, preventing STDEVCLIP from returning 0 (fix suggested by Robert Lupton).
  - Add sigma68 fallback: if STDEVCLIP still returns 0, use (p84 - p16) / 2 as a safety net.
  - Add configurable `nPixMin` parameter for footprint detection (previously hardcoded to 10).

  ## Problem

  `STDEVCLIP` (from `lsst.afw.math`) returns 0 on some guider coadd images, causing
  `FootprintSet` to crash with `InvalidParameterError: 'St. dev. must be > 0: 0'`.

  Root cause: the guider coadd is a median stack of ~130 stamps. This averages the noise down
  so much that the pixel distribution becomes extremely narrow — many pixels land in the same
  quantized bin. When STDEVCLIP iterates, the first-pass sigma is near zero, the clipping
  range collapses, and it converges to std=0.

  ## Validation

  Full validation on night 2026-02-22: 447 sequences (255 previously failing + 192 passing
  control), excluding 24 with DatasetNotFoundError.

  **STDEVCLIP fix:** The dither fix eliminates all STDEVCLIP=0 cases. Before: 437/2040 (21.4%)
   guider measurements hit STDEVCLIP=0 on failing images, and 25/1536 (1.6%) on passing
  images. After: 0 across all 3576 measurements. The sigma68 fallback was never triggered,
  confirming the dither alone is sufficient.

  **Star recovery:** On the 255 previously failing sequences, the new code recovers a median
  of 7 stars per sequence (was 0). On the 192 control sequences, star counts are stable: 153
  unchanged, 32 improved, 7 lose 1–2 stars due to GalSim `FindAdaptiveMom` convergence
  failures (not related to this fix, tracked in DM-54305).

  | Stage | Stars | Possible | Rate |
  |-------|-------|----------|------|
  | Old pipeline | 1345 | 3576 | 37.6% |
  | + DM-54263 (this PR) | 3144 | 3576 | 87.9% |

### DIFF
--- a/python/lsst/summit/utils/guiders/detection.py
+++ b/python/lsst/summit/utils/guiders/detection.py
@@ -277,7 +277,7 @@ def runSourceDetection(
     cutOutSize: int = 25,
     apertureRadius: int = 5,
     gain: float = 1.0,
-    starMaskThreshold: float = 20000.0,
+    nPixMin: int = 10,
 ) -> pd.DataFrame:
     """
     Detect sources in an image and measure their properties.
@@ -294,9 +294,8 @@ def runSourceDetection(
         Aperture radius in pixels for photometry.
     gain : `float`
         Detector gain (e-/ADU).
-    starMaskThreshold : `float`
-        Pixel count threshold for masking bright stars when computing the
-        background standard deviation. Pixels above this value are excluded.
+    nPixMin : `int`
+        Minimum number of pixels in a footprint for detection.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


python/lsst/summit/utils/guiders/detection.py
```diff
@@ -303,12 +307,30 @@ def runSourceDetection(
...
+            )
+        imageStd = np.std(bgPixels)
+        if imageStd <= 0:
+            raise ValueError(
+                f"Background standard deviation is {imageStd}. "
+                "Image may be flat or starMaskThreshold may need adjustment."
+            )
+        absThreshold = threshold * imageStd
+        thresh = afwDetect.Threshold(absThreshold, afwDetect.Threshold.VALUE)
+        footprints = afwDetect.FootprintSet(exposure.getMaskedImage(), thresh, "DETECTED", 10)
```
mfisherlevine: I don't love hardcoding the min number of pixels to 10, it might be worth making it possible to pass that in.

estevesjh: Got it. I will make this change.

estevesjh: Done, now nPixMin is an arg.

TallJimbo (approved):
Changes look ok.  I'm surprised to see algorithmic code in `summit_utils` at all, but that's not an issue for this ticket.

PR #190, opened 2026-06-03
OSW-2391: Add support in ConsDbClient to insert by (day_obs, seq_num, detector)

bbrondel:
This change updates the insert() method for ConsDB client. It changes the method to allow a tuple of 3 integers containing day_obs, seq_num, detector. This allows the method to be used to insert into ccd-level tables using the multi-column primary keys.

### DIFF
--- a/python/lsst/summit/utils/consdbClient.py
+++ b/python/lsst/summit/utils/consdbClient.py
@@ -481,7 +481,7 @@ class ConsDbClient:
         self,
         instrument: str,
         table: str,
-        obs_id: tuple[int, int] | int,
+        obs_id: tuple[int, int] | tuple[int, int, int] | int,
         values: Mapping[str, Any],
         *,
         allow_update: bool = False,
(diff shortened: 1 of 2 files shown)


python/lsst/summit/utils/consdbClient.py
```diff
@@ -526,15 +536,21 @@ def insert(
 
         data: dict[str, Any]
         if isinstance(obs_id, tuple):
+            if len(obs_id) not in (2, 3):
```
pkubanek: That can be better (a bit shorter, and also assert signals better the intend) written as:

assert len(obs_id) in (2, 3), f"obs_id tuple must be (day_obs, seq_num) or (day_obs, seq_num, detector); got {obs_id!r}"

pkubanek: I assume tuple length points to internal error.. 

Image

pkubanek (approved):
Looks good, see comments.

PR #192, opened 2026-06-04
DM-55152: Dedupe columns in getCcdVisitTableForDay

mfisherlevine:
cvq.* already selects every column of ccdvisit1_quicklook, so naming detector/visit_id/seq_num/etc. again explicitly from the joined ccdvisit1 and visit1 tables produces duplicate column names once ConsDB starts denormalising those identity columns onto the quicklook table. astropy refuses to build a Table from a result set with repeated column names, raising "ValueError: Duplicate column names".

Skip any explicitly-named column the quicklook schema already provides, mirroring the dedup getWideQuicklookTableForDay already does, and qualify the optional detector filter as cv.detector so it stays unambiguous.

Add regression tests covering the overlapping and non-overlapping column cases.

### DIFF
--- a/python/lsst/summit/utils/consdbClient.py
+++ b/python/lsst/summit/utils/consdbClient.py
@@ -740,18 +740,40 @@ def getCcdVisitTableForDay(
     table : `astropy.table.Table`
         The resulting table.
     """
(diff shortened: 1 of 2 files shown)


bbrondel (approved):
This looks great, thank you for your efforts on this.

PR #193, opened 2026-07-24
DM-55094: Add package version data model and its ConsDB round-trip

mfisherlevine:
Rapid analysis records, for every image it dispatches, the versions of the science packages that determine the AOS results, plus an overall incrementing version number mapped in a small per-instrument on-disk registry. The data model for that — the PackageVersions dataclass, its JSON serialisation, and the registry read/write — moves here from rubintv_production so that users without access to that package can consume the data.

ConsDB is about to grow a JSONB column for this. writePackageVersionsToConsDb and readPackageVersionsFromConsDb round-trip the pinned blob shape ({"versions": {...}, "version_number": N}) through it, defaulting to cdb_<instrument>.exposure_quicklook.package_versions until the schema change lands (table and column are parameterised in case it lands elsewhere). loadPackageVersionRegistry and writeBackdatedPackageVersions support back-filling ConsDB for images processed before the column existed, expanding a known overall version number to its full package-version set via the registry.

The blob shape and the deployed registry file format are pinned by verbatim-decode unit tests, and the ConsDB read/write paths are tested against a mocked server via responses, including the write -> read round-trip.

### DIFF
--- a/python/lsst/summit/utils/packageVersions.py
+++ b/python/lsst/summit/utils/packageVersions.py
@@ -1,0 +1,633 @@
+# This file is part of summit_utils.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""The science-package version data model, and its ConsDB round-trip.
+
+The Rapid Analysis backend records, for every image it dispatches, the exact
+versions of the handful of science packages that determine the AOS results,
+plus a single overall "version number" - an integer that increments whenever
+the combined set of package versions changes. The mapping from number to
+versions is kept in a small per-instrument JSON registry file, owned and
+written solely by Rapid Analysis.
+
+This module is the public home of that data model. It lives here, rather than
+in the Rapid Analysis package, because most users cannot import Rapid Analysis
+code but do need to consume the data: `PackageVersions` and its
+(de)serialisation define the wire format, and the functions below read and
+write it in its three homes:
+
+- the ConsDB ``package_versions`` JSONB column (pending: see
+  `PACKAGE_VERSIONS_TABLE`), via `writePackageVersionsToConsDb` /
+  `readPackageVersionsFromConsDb`;
+- the on-disk version-number registry, via `resolveVersionNumber` (Rapid
+  Analysis' write path) and `loadPackageVersionRegistry` (everyone else's read
+  path);
+- and, for data taken before the ConsDB column existed,
+  `writeBackdatedPackageVersions` back-fills ConsDB from the registry.
+
+The JSON blob written to ConsDB is::
+
+    {"versions": {"ts_wep": "v17.6.1-alpha", ...}, "version_number": 42}
+
+with ``version_number`` null when it could not be determined. The blob shape is
+pinned by tests in both this package and Rapid Analysis; change it only with a
+migration plan for the rows already written.
+"""
+
+from __future__ import annotations
+
+__all__ = [
+    "UNKNOWN_VERSION",
+    "UNKNOWN_VERSION_NUMBER",
+    "PACKAGE_VERSIONS_TABLE",
+    "PACKAGE_VERSIONS_COLUMN",
+    "PackageVersions",
+    "getRegistryPath",
+    "loadPackageVersionRegistry",
+    "resolveVersionNumber",
+    "writePackageVersionsToConsDb",
+    "readPackageVersionsFromConsDb",
+    "writeBackdatedPackageVersions",
+]
+
+import datetime
+import hashlib
+import json
+import logging
+import os
+import re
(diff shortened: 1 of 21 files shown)


python/lsst/summit/utils/packageVersions.py
```diff
@@ -0,0 +1,427 @@
...
+        blobDict : `dict` [`str`, `Any`]
+            The pinned wire shape: ``{"hash": ..., "versions": {...}}``.
+            ``versions`` is inline so the blob is self-contained; ``hash`` is
+            included as a stable SQL-side identity/dedup key.
+        """
+        return {"hash": self.versionHash(), "versions": dict(self.versions)}
+
+    @classmethod
+    def fromDict(cls, blobDict: Mapping[str, Any]) -> PackageVersions:
+        """Build from the JSON blob dict stored in the ConsDB column.
```
ctslater: A dict is not a blob, to my thinking.

tests/test_packageVersions.py
```diff
@@ -0,0 +1,252 @@
...
+    assert readPackageVersionsFromConsDb(client, "LSSTCam", 20250624, 123) is None
+
+
+@responses.activate
+def test_readPackageVersionsForExposure(client: ConsDbClient) -> None:
+    # the DimensionRecord wrapper must unpack the dataId and delegate: it
+    # produces the same query keyed by this record's day_obs/seq_num
+    blob = {"hash": EXPECTED_HASH, "versions": VERSIONS}
+    expectedQuery = (
+        "SELECT q.package_versions "
```
ctslater: This is effectively just a test of string substitution if you have to include basically the same string that's in the test. I don't think it's worth it (same on the three other tests).

tests/test_packageVersions.py
```diff
@@ -0,0 +1,252 @@
...
+    # the JSONB column coming back as an already-parsed JSON object; the query
+    # SQL is pinned here too (the join via the exposure table)
+    blob = {"hash": EXPECTED_HASH, "versions": VERSIONS}
+    expectedQuery = (
+        "SELECT q.package_versions "
+        "FROM cdb_lsstcam.exposure_quicklook q "
+        "JOIN cdb_lsstcam.exposure e ON e.exposure_id = q.exposure_id "
+        "WHERE e.day_obs = 20250624 AND e.seq_num = 123"
+    )
+    responses.post(
```
ctslater: I could use a comment saying this is setting up a mock (unfortunately non-self-explanatory name for this library...)

tests/test_packageVersions.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+@responses.activate
+def test_readPackageVersionsFromConsDbParsesJsonObjectCell(
+    client: ConsDbClient,
+) -> None:
+    # the JSONB column coming back as an already-parsed JSON object; the query
+    # SQL is pinned here too (the join via the exposure table)
+    blob = {"hash": EXPECTED_HASH, "versions": VERSIONS}
+    expectedQuery = (
```
ctslater: This is effectively just a test of string substitution if you have to include basically the same string that's in the test. I don't think it's worth it (same on the three other tests).

tests/test_packageVersions.py
```diff
@@ -0,0 +1,252 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Tests for the science-package version data model and reading it from ConsDB.
+
+The ConsDB JSON blob (``{"hash": ..., "versions": {...}}``) is a wire format
```
ctslater: This comment is just fluff.

python/lsst/summit/utils/packageVersions.py
```diff
@@ -0,0 +1,427 @@
...
+    Tolerant of how the JSONB column is delivered over the query API: a null
+    (or masked) cell yields `None`, an already-parsed JSON object or a JSON
+    string both yield a `PackageVersions`.
+
+    Parameters
+    ----------
+    cell : `Any`
+        The raw cell value from the query result.
+    column : `str`
+        The column name, for the error message only.
```
ctslater: Shouldn't take unnecessary arguments. If the caller of the function wants to report the error with additional info they can catch and re-raise.

python/lsst/summit/utils/packageVersions.py
```diff
@@ -0,0 +1,427 @@
...
+        f"WHERE q.{column}->>'hash' = '{versionHash}' "
+        "LIMIT 1"
+    )
+    result = client.query(query)
+    if len(result) == 0:
+        return None
+    return _packageVersionsFromCell(result[column][0], column)
+
+
+def _packageVersionsFromCell(cell: Any, column: str) -> PackageVersions | None:
```
ctslater: Why is this called a cell?

python/lsst/summit/utils/packageVersions.py
```diff
@@ -0,0 +1,427 @@
...
+    query = (
+        f"SELECT q.{column} "
+        f"FROM {schemaName}.{table} q "
+        f"WHERE q.{column}->>'hash' = '{versionHash}' "
+        "LIMIT 1"
+    )
+    result = client.query(query)
+    if len(result) == 0:
+        return None
+    return _packageVersionsFromCell(result[column][0], column)
```
ctslater: Is that extra function call better than `return PackageVersions.fromJson(result[column][0])`?

ctslater (approved):

tests/test_packageVersions.py
mfisherlevine: Dropped them. Kept a check that the hash actually gets lowercased before it hits the query in the one test, since that's the whole bug it's there for, but no more matching the full SQL.

python/lsst/summit/utils/packageVersions.py
mfisherlevine: It's doing a bit more than that - null/masked handling, and the column doesn't exist in consdb yet so I don't actually know whether the JSONB comes back as a dict or a string. Once it lands and we know, happy to collapse it to a one-liner.

